PR #99, opened 2020-01-31
[WIP] Implemented Thanos

K-DOT:


teodor-pripoae:
LGTM

pkg/phases/monitoring/monitoring.go
```diff
@@ -41,6 +41,26 @@ func Install(p *platform.Platform) error {
...
+		exists, err := s3Client.BucketExists(p.Thanos.Bucket)
+		if err != nil {
+			return err
+		}
+		if !exists {
+			if err := s3Client.MakeBucket(p.Thanos.Bucket, p.S3.Region); err != nil {
+				return err
+			}
+		}
+		p.ApplySpecs("", "monitoring/thanosConfig.yaml")
```
moshloop:
```suggestion
		if err := p.ApplySpecs("", "monitoring/thanosConfig.yaml"); err != nil {
		      return err
		}
```

pkg/types/types.go
```diff
@@ -336,6 +337,12 @@ type CA struct {
 	Password   string `yaml:"password,omitempty"`
 }
 
+type Thanos struct {
+	Disabled bool   `yaml:"disabled,omitempty"`
+	Version  string `yaml:"tag,omitempty"`
```
moshloop:
```suggestion
	Version  string `yaml:"version,omitempty"`
```

manifests/monitoring/observability/thanosCompactor.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
moshloop: Why does this need to be a statefulset ?

pkg/phases/monitoring/monitoring.go
```diff
@@ -41,6 +42,46 @@ func Install(p *platform.Platform) error {
 		return err
 	}
 
+	if p.Thanos == nil || p.Thanos.Disabled {
+		log.Println("Thanos is disabled")
```
moshloop:
```suggestion
		log.Debugln("Thanos is disabled")
```

pkg/types/types.go
```diff
@@ -344,6 +345,15 @@ type CA struct {
 	Password   string `yaml:"password,omitempty"`
 }
 
+type Thanos struct {
+	Disabled               bool   `yaml:"disabled,omitempty"`
+	Version                string `yaml:"version,omitempty"`
+	Mode 	               string `yaml:"mode,omitempty"`
+	ThanosSidecarEndpoint  string `yaml:"thanosSidecarEndpoint,omitempty"`
+	ThanosSidecarPort      string `yaml:"thanosSidecarPort,omitempty"`
+	S3                    *S3     `yaml:"s3,omitempty"`
```
moshloop: I think this should be a bucket name only and then we piggy-back off platform.S3 for credentials

manifests/monitoring/observability/thanosCompactor.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: apps/v1
```
moshloop: `s/thanosCompactor.yaml/thanos-compactor.yml/` as it is more consistent with existing naming conventions

manifests/monitoring/kube-prometheus.yml
```diff
@@ -133,6 +141,7 @@ spec:
       port: 9090
       targetPort: web
   selector:
+    app: prometheus
```
moshloop: Is this a duplicate line ?

pkg/phases/monitoring/monitoring.go
```diff
@@ -41,6 +42,46 @@ func Install(p *platform.Platform) error {
...
+		}
+		thanosSpecs := []string{"thanosCompactor.yaml",  "thanosQuerier.yaml", "thanosStore.yaml"}
+		for _, spec := range thanosSpecs {
+			log.Infof("Applying %s", spec)
+			if err := p.ApplySpecs("", "monitoring/observability/"+spec); err != nil {
+				return err
+			}
+		}
+	} else {
+		log.Printf("Thanos: wrong mode %s. Should be client or observability", p.Thanos.Mode)
```
moshloop: fail-fast and hard on config errors.

manifests/monitoring/thanosConfig.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if not (index .thanos "disabled") }}
```
moshloop: is this if statement required?

manifests/monitoring/observability/thanosCompactor.yaml
K-DOT: Because Compactor by itself is not concurrency safe:
https://github.com/thanos-io/thanos/blob/master/docs/components/compact.md

pkg/phases/monitoring/monitoring.go
K-DOT: Will change

pkg/types/types.go
K-DOT: There can't be multiple buckets for Thanos Querier. So we can't use local Minio for storing metrics for every client cluster. Or platform.S3 is the same for all clusters?

manifests/monitoring/observability/thanosCompactor.yaml
K-DOT: Sure

manifests/monitoring/kube-prometheus.yml
K-DOT: What do you mean? Looks correct.

pkg/phases/monitoring/monitoring.go
K-DOT: Ok, will change

manifests/monitoring/thanosConfig.yaml
K-DOT: Not required, but without it will have Thanos configmap even if we don't have Thanos itself.

manifests/monitoring/kube-prometheus.yml
moshloop: Line 144 and 145 are the same?

manifests/monitoring/observability/thanosCompactor.yaml
moshloop: I think a Deployment with a maxSurge: 0, and maxUnavailable: 1 is a better way to model this constraint.

pkg/types/types.go
moshloop: Yes, platform.S3 should point to the same object store (minio / S3) for all clusters.

manifests/monitoring/thanosConfig.yaml
moshloop: Is the `ApplySpec('monitoring/thano-config/yaml')` not guarded by an If statement in install.go ?

manifests/monitoring/kube-prometheus.yml
```diff
@@ -119,6 +119,55 @@ spec:
...
+        valueFrom:
+          fieldRef:
+            apiVersion: v1
+            fieldPath: status.podIP
+      - name: OBJSTORE_CONFIG
+        valueFrom:
+          secretKeyRef:
+            key: thanos.yaml
+            name: thanos-objstore-config
+      image: thanosio/thanos:v0.10.1
```
moshloop:
```suggestion
      image: thanosio/thanos:{{.thanos.version}}
```

pkg/phases/monitoring/monitoring.go
```diff
@@ -41,6 +46,80 @@ func Install(p *platform.Platform) error {
...
+		}
+		thanosSpecs := []string{"thanos-compactor.yaml", "thanos-querier.yaml", "thanos-store.yaml"}
+		for _, spec := range thanosSpecs {
+			log.Infof("Applying %s", spec)
+			if err := p.ApplySpecs("", "monitoring/observability/"+spec); err != nil {
+				return err
+			}
+		}
+	} else {
+		log.Fatalf("Thanos: wrong mode %s. Should be client or observability", p.Thanos.Mode)
```
moshloop:
```suggestion
		return fmt.Errorf("invalid thanos mode '%s',  valid options are  'client' or 'observability'", p.Thanos.Mode)
```

pkg/types/types.go
```diff
@@ -277,6 +283,16 @@ type CA struct {
 	Password   string `yaml:"password,omitempty"`
 }
 
+type Thanos struct {
+	Disabled              bool     `yaml:"disabled,omitempty"`
+	Version               string   `yaml:"version,omitempty"`
```
moshloop: For required fields, remote the omitempty
For other fields add comments to each field to descript its purpose and possible options so that they show up in the docs.

manifests/monitoring/kube-prometheus.yml
```diff
@@ -105,7 +105,7 @@ spec:
   resources:
     requests:
       memory: 400Mi
-  retention: 240h
```
moshloop: Does this revert to the default retention period ?

PR #104, opened 2020-02-05
Fix NodeLocalDNS

mazzy89:
This PR aims to fix NodeLocalDNS.

It is still not tested.

I need to verify how the following values:

- `__PILLAR__CLUSTER__DNS__`
- `__PILLAR__UPSTREAM__SERVERS__`

are changed.

According to the documentation

> The following variables will be set by the node-cache images - k8s.gcr.io/k8s-dns-node-cache:1.15.6 or later. The values will be determined by reading the kube-dns configMap for custom Upstream server configuration.
`__PILLAR__CLUSTER__DNS__` - Upstream server for in-cluster queries.
`__PILLAR__UPSTREAM__SERVERS__` - Upstream servers for external queries.

But still not clear to me how.

### DIFF
--- a/manifests/node-local-dns.yml
+++ b/manifests/node-local-dns.yml
@@ -54,54 +54,54 @@ metadata:
     addonmanager.kubernetes.io/mode: Reconcile
 data:
   Corefile: |
-    __PILLAR__DNS__DOMAIN__:53 {
+    {{ .nodeLocalDNS.dnsDomain }}:53 {
         errors
         cache {
-                success 9984 30
-                denial 9984 5
+            success 9984 30
+            denial 9984 5
         }
         reload
         loop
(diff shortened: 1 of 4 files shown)


mazzy89:
Allright according the author's PR this seems really magic and node-local-dns will do the rest https://github.com/kubernetes/kubernetes/pull/82845#commitcomment-35529267

mazzy89:
Actually, in the near future, we could even remove the logic to get `kube-dns` Service IP. Indeed

> Autofilling the DNS Server is possible, node-cache code just needs to look for the kube-dns service ip by checking the value of env variable - KUBE_DNS_SERVICE_HOST

This feature is not implemented yet but definitely will make neater the implementation here.

PR #108, opened 2020-02-08
Improved logging

K-DOT:


moshloop:
This is my fault for not highlighting this convention earlier: https://github.com/flanksource/style-guide/blob/master/GOLANG.md#wrap-errors-rather-than-logging-and-returning 

But we want to wrap and return errors, and not log and return

PR #117, opened 2020-02-13
Added Contour as an alternative for nginx ingress

### LINKED ISSUE #22: Contour
https://projectcontour.io/ as an alternative to nginx.

K-DOT:
Fixes #22


moshloop:
Can you common on the PR: "Fixes #22" so that when this PR merges, the issue automatically gets closed ?

Can we also rather vendor the YAML into manifests rather than using the helm chart ? 
https://raw.githubusercontent.com/projectcontour/contour/release-1.1/examples/render/contour.yaml

K-DOT:
> Can you common on the PR: "Fixes #22" so that when this PR merges, the issue automatically gets closed ?
> 
> Can we also rather vendor the YAML into manifests rather than using the helm chart ?
> https://raw.githubusercontent.com/projectcontour/contour/release-1.1/examples/render/contour.yaml

Sure

pkg/phases/base/install.go
```diff
@@ -95,11 +96,16 @@ func Install(platform *platform.Platform) error {
...
 	}
 
-	if platform.Nginx == nil || !platform.Nginx.Disabled {
+	if platform.Nginx != nil && !platform.Nginx.Disabled {
 		log.Infof("Installing Nginx Ingress Controller")
 		if err := platform.ApplySpecs("", "nginx.yml"); err != nil {
 			log.Errorf("Error deploying nginx: %s\n", err)
 		}
+	} else if platform.Contour != nil && !platform.Contour.Disabled {
+		log.Infof("Skipping deployment of nginx. Contour will be deployed")
```
moshloop:
```suggestion
		log.Infof("Installing Contour Ingress Controller")
```

pkg/phases/base/install.go
```diff
@@ -95,11 +96,16 @@ func Install(platform *platform.Platform) error {
 		}
 	}
 
-	if platform.Nginx == nil || !platform.Nginx.Disabled {
+	if platform.Nginx != nil && !platform.Nginx.Disabled {
```
moshloop:
```suggestion
        // Deploy nginx ingress controller if neither nginx or contour is specified
	if (platform.Nginx == nil && platform.Contour == nil) || (platform.Nginx != nil && !platform.Nginx.Disabled) {
```

moshloop: Perhaps wrap the if statement into a `platform.GetIngressController` for reuse

pkg/phases/base/test.go
```diff
@@ -19,12 +20,12 @@ func Test(platform *platform.Platform, test *console.TestResults) {
...
-	if platform.Nginx == nil || !platform.Nginx.Disabled {
+	if (platform.Nginx == nil || !platform.Nginx.Disabled) && (platform.Contour == nil || platform.Contour.Disabled )  {
 		k8s.TestNamespace(client, "ingress-nginx", test)
 	}
 
 	if platform.Minio == nil || !platform.Minio.Disabled {
 		k8s.TestNamespace(client, "minio", test)
 	}
-
+	contour.Test(platform, test)
```
moshloop: this should be tested inline e.g.

```golang
if platform.GetIngressController() == "contour") {
    k8s.TestNamespace(client, "contour")
} else {
    k8s.TestNamespace(client, "nginx-ingress")
}
```
The tests would only be considered a skip if both nginx and contour are disabled.

test/common.yml
```diff
@@ -80,3 +79,6 @@ fluentd-operator:
   repository: jvassev/kube-fluentd-operator
 eck:
   version: 1.0.0
+contour:
+  version: v0.15.0
+  chartVersion: 0.2.0
```
moshloop:
```suggestion
```

test/e2e.sh
```diff
@@ -43,41 +43,41 @@ $BIN deploy stubs -v
 
 $BIN test base --wait 200
 
-$BIN deploy pgo install -v
+# $BIN deploy pgo install -v
```
moshloop: I think these were inadvertently committed

go.mod
```diff
@@ -6,13 +6,15 @@ require (
 	cloud.google.com/go v0.47.0 // indirect
 	cloud.google.com/go/bigquery v1.2.0 // indirect
 	cloud.google.com/go/storage v1.2.1 // indirect
+	github.com/alecthomas/kingpin v2.2.6+incompatible
```
moshloop: Don't bump module versions unnecessarily - if they need to be bumped, it is normally better todo it in a new PR

moshloop:
It looks like you may have made some of these commits with a different git config user.name or user.email which is why the CLA is failing.

go.mod
K-DOT: This new dependency was added automatically. I don't know where is it used.

K-DOT:
> It looks like you may have made some of these commits with a different git config user.name or user.email which is why the CLA is failing.
Yes, I changed my laptop and it was set as default. Can we fix it?

teodor-pripoae:
@K-DOT  You can rebase with master and create one new commit instead of the 5.

teodor-pripoae:
It will fix the email

PR #103, opened 2020-02-04
Add command to provision kind cluster #36

teodor-pripoae:
Screenshot 2020-02-04 12 20 04
Screenshot 2020-02-04 12 20 14

PR #118, opened 2020-02-15
Issue 84 test openldap

teodor-pripoae:
This is based on #103 PR so we need to first merge #103.


manifests/dex.cfg
```diff
@@ -21,10 +21,10 @@ connectors:
   id: ldap
   name: LDAP
   config:
-    host: {{.ldap.host}}:636
+    host: {{.ldap.host}}:{{.ldap.port}}
     insecureSkipVerify: true
-    bindDN: $AD_USERNAME
-    bindPW: $AD_PASSWORD
+    bindDN: {{.ldap.bindDN}}
```
moshloop: We don't want to inject the username and password directly into the configmap as they are sensitive, we, therefore, store them in secrets, load them into DEX as environment variables where DEX does the replacement.

pkg/types/types.go
```diff
@@ -216,21 +215,22 @@ type NFS struct {
...
+	Disabled   bool   `yaml:"disabled,omitempty"`
 	Host       string `yaml:"host,omitempty"`
+	Port       string `yaml:"port,omitempty"`
 	Username   string `yaml:"username,omitempty"`
 	Password   string `yaml:"password,omitempty"`
 	Domain     string `yaml:"domain,omitempty"`
 	AdminGroup string `yaml:"adminGroup,omitempty"`
-	BindDN     string `yaml:"dn,omitempty"`
+	DN         string `yaml:"dn,omitempty"`
+	BindDN     string `yaml:"bindDN,omitempty"`
```
moshloop: how do username and BindDN differ?

teodor-pripoae: Username is used also for login in dex. It should be something like `test`.  BindDN is used for ldap first verification, something like `admin` with password `secret`.

This is what gets generated:

```yaml
- type: ldap
  id: ldap
  name: LDAP
  config:
    host: apacheds.ldap:10636
    insecureSkipVerify: true
    bindDN: uid=admin,ou=system
    bindPW: secret

    userSearch:
      baseDN: ou=users,dc=example,dc=com
      filter: "(objectClass=person)"
      username: sAMAccountName
      idAttr: DN
      emailAttr: mail
      nameAttr: displayName

    groupSearch:
      baseDN: ou=users,dc=example,dc=com
      filter: "(objectClass=group)"
      userAttr: DN
      groupAttr: member
      nameAttr: name
```

This is the config:

```yml
ldap:
  adminGroup: NA1
  username: test
  password: secret
  port: 10636
  host: apacheds.ldap
  dn: ou=users,dc=example,dc=com
  bindDN: uid=admin,ou=system
  bindPW: secret
```

teodor-pripoae: I've changed it back to defaults.

pkg/phases/dex/test.go
```diff
@@ -1,12 +1,64 @@
...
+		RedirectURI:  "http://localhost:8000",
+		Username:     "test",
+		Password:     "secret",
+	}
+
+	token, err := dexClient.GetAccessToken()
+	if err != nil {
+		test.Failf("dex", "failed to get token %v", err)
+		return
+	}
```
moshloop: I think add in a `test.Passf("dex", "OIDC Authentication flow")`

pkg/phases/dex/test.go
```diff
@@ -1,12 +1,64 @@
...
+	if err != nil {
+		test.Failf("dex", "failed to create k8s config: %v", err)
+		return
+	}
+	pods, err := k8s.CoreV1().Pods("dex").List(metav1.ListOptions{})
+	if err != nil {
+		test.Failf("dex", "failed to list pods: %v", err)
+		return
+	}
+	for _, pod := range pods.Items {
```
moshloop: add a `test.Passf("dex", "OIDC Kubeconfig access")

PR #131, opened 2020-02-18
Fix reading ca certificate when password is empty

teodor-pripoae:


pkg/platform/platform.go
```diff
@@ -82,6 +82,9 @@ func (platform *Platform) GetCA() certs.CertificateAuthority {
 func readCA(ca *types.CA) (*certs.Certificate, error) {
 	cert := files.SafeRead(ca.Cert)
 	privateKey := files.SafeRead(ca.PrivateKey)
+	if ca.Password == "" {
```
moshloop: This should probably go into certs.DecryptCertificate that attempts to just decode the cert if the password is empty

moshloop: This should probably log a warning as well.

teodor-pripoae: Then I should change it inside `flanksource/commons`

PR #143, opened 2020-02-27
reduce usage of kubectl external command

moshloop:


mazzy89:
I'll take a look at it later today.

PR #161, opened 2020-03-07
Added PushGateway and implemented e2e checks for Thanos

### LINKED ISSUE #146: [E2E] Thanos
Ensure Thanos is healthy and is aggregating metrics from 2 clusters.

as a parameter to the test take 2 kubecontexts a client and an observability mode cluster that have already been setuo

K-DOT:
Fixes #146


go.mod
```diff
@@ -3,45 +3,54 @@ module github.com/moshloop/platform-cli
 go 1.12
 
 require (
+	github.com/alecthomas/kingpin v2.2.6+incompatible
 	github.com/aws/aws-sdk-go v1.29.5
-	github.com/davecgh/go-spew v1.1.1
```
moshloop: Go-spew is used, and crunchydata isn't ?

cmd/test.go
```diff
@@ -177,6 +195,17 @@ func init() {
 		},
 	})
 
+
+	Test.AddCommand(&cobra.Command{
+		Use:   "thanos",
+		Short: "Test thanos. Requires Pushgateway and Thanos endpoints",
+		Long: "test thanos [pushgateway] [thanos]",
```
moshloop: I think make this named arguments e.g. `--pushgateway` or `--thanos` and use the local instances if not specified

manifests/monitoring/pushgateway.yaml
```diff
@@ -0,0 +1,60 @@
...
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  labels:
+    k8s-app: pushgateway
+  name: pushgateway
+  namespace: monitoring
+spec:
+  rules:
+    - host: pushgateway.{{.name}}.lab.flanksource.com
```
moshloop: This should be `pushgateway.{{.domain}}`

pkg/phases/monitoring/test.go
```diff
@@ -1,12 +1,94 @@
...
+	thanosHost := args[1]
+	pusher := pushMetric(pushGatewayHost)
+	err := pusher.Add()
+    if err != nil {
+		test.Failf("Thanos: client", "Failed to inject metric to client Prometheus via Pushgateway %v", err)
+		return
+	} else {
+		test.Passf("Thanos client", "Metric successfully injected into the client Prometheus. Waiting to receive it in observability cluster.")
+	}
+	log.Tracef("Waiting for metric")
```
moshloop: Put this into a loop that does the check, and sleeps for 5 seconds up until the wait timeout so that we can exit early

pkg/phases/monitoring/test.go
```diff
@@ -1,12 +1,94 @@
...
+}
+
+func pushMetric(pushGatewayHost string) *push.Pusher {
+	metric := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: testMetricName,
+		Help: "This metric is created from Thanos e2e tests. You can ignore it.",
+	})
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(metric)
+	metric.Add(1)
```
moshloop: I think this should probably be the current epoch  to prevent returning true on stale values.

go.mod
K-DOT: It should be a mistake. Will change.

cmd/test.go
K-DOT: Ok, will do

pkg/phases/monitoring/test.go
K-DOT: I implemented it this way first, but don't we have retry logic already?

pkg/phases/monitoring/test.go
K-DOT: There will be no stale values because we are deleting metric after fist successful run. But I will double-check.

manifests/monitoring/pushgateway.yaml
K-DOT: > Can we also a short description to the command on how the test works ?

Sure

PR #172, opened 2020-03-08
Added configmap-reloader deployment

K-DOT:


moshloop:
Can we add docs and an e2e test for this ?

Issue #162, opened 2020-03-08
Support importing config files


moshloop:
e.g. instead of `-c config.yml -c common.yml`

`config.yml`
```yaml
importConfigs:
    - common.yml
```

PR #179, opened 2020-03-09
Import other configs from config.yml #162

teodor-pripoae:


moshloop:
Can we add some docs - and maybe a unit test or 2, specifically around import priority / ordering ?

PR #188, opened 2020-03-10
Issue 170 e2e git ops

teodor-pripoae:


pkg/phases/flux/flux.go
```diff
@@ -112,7 +116,7 @@ func NewFluxDeployment(cr *types.GitOps) []runtime.Object {
 	if cr.Namespace == "kube-system" {
 		spec.ServiceAccount(saName).AddClusterRole("cluster-admin")
 	} else {
-		spec.ServiceAccount(saName).AddRole("cluster-admin")
+		spec.ServiceAccount(saName).AddClusterRole("cluster-admin")
```
moshloop: I think this should be AddRole("namespace-admin")

test/e2e.sh
```diff
@@ -71,6 +71,8 @@ $BIN deploy eck
 
 $BIN deploy opa policies test/opa/policies -v
 
+$BIN deploy gitops
```
moshloop: This should not be necessary, as `deploy all` should also deploy gitops controllers

pkg/phases/flux/flux.go
```diff
@@ -112,7 +116,7 @@ func NewFluxDeployment(cr *types.GitOps) []runtime.Object {
 	if cr.Namespace == "kube-system" {
 		spec.ServiceAccount(saName).AddClusterRole("cluster-admin")
 	} else {
-		spec.ServiceAccount(saName).AddRole("cluster-admin")
+		spec.ServiceAccount(saName).AddClusterRole("namespace-admin").AddClusterRole("namespace-creator")
```
moshloop:
```suggestion
		spec.ServiceAccount(saName).AddRole("namespace-admin").AddRole("namespace-creator")
```

PR #206, opened 2020-03-19
Documentation enhancements

philipstaffordwood:
Some documentation enhancements, mostly around developer instructions.
* Added a `help` target to the `makefile` which displays valid targets and basic usage,
* Added some "placeholder" pages to prevent 404's on the live site,
* Added some developer sections for quickstart and documentation viewing/editing


.gitignore
```diff
@@ -6,7 +6,8 @@ build/
 *.env
 config*.yaml
 config*.yml
-docs/*.md
+#TODO: verify intent of ignoring docs/*.md at PR
+#docs/*.md
```
philipstaffordwood: Is removing `docs/*.md` from the `.gitignore`, OK?
I could quite understand why we are doing it...

I'm allowing it in this PR to add a placeholder page to prevent the http 404 on this link:
https://platform-cli.docs.flanksource.com/

moshloop: It should just be to ignore the generated docs - primarily reference/config.md

PR #233, opened 2020-03-30
Adding k8s auditing support for KIND clusters

philipstaffordwood:
This PR for https://github.com/flanksource/platform-cli/issues/57 adds logfile auditing support for KIND clusters.


pkg/k8s/client.go
```diff
@@ -1005,3 +1005,46 @@ func NewCommandJob(node, command string) v1.PodSpec {
 		HostIPC:     true,
 	}
 }
+
+// Returns the name of first node found labeled as a master
```
moshloop:
```suggestion
// GetMasterNode returns the name of the first node found labelled as a master
```

pkg/types/config.go
```diff
@@ -1,6 +1,7 @@
 package types
 
 type PlatformConfig struct {
+	AuditConfig           *AuditConfig      `yaml:"auditConfig,omitempty"`
```
moshloop: I think this should be moved into the Kubernetes object

pkg/types/types.go
```diff
@@ -517,6 +520,54 @@ type Connection struct {
 	Verify   string `yaml:"verify,omitempty"`
 }
 
+type AuditConfig struct {
+	Disabled         bool             `yaml:"disabled,omitempty"`
+	PolicyFile       string           `yaml:"auditPolicyFile,omitempty"`
+	APIServerOptions APIServerOptions `yaml:"kubeApiServerOptions,omitempty"`
```
moshloop: I don't think we need this -we can use sane defaults, and then use the apiServer.extraArgs to override: 

e.g.

```yaml
kubernetes:
    auditing:
       policyFile: ./policy.yaml
    apiServerExtraArgs:
       "audit-log-maxsize": 1GB
```

pkg/phases/audit/test.go
```diff
@@ -0,0 +1,144 @@
...
+
+//Test k8s auditing functionality.
+func Test(p *platform.Platform, tr *console.TestResults) {
+	ac := p.AuditConfig
+
+	if ac == nil {
+		// A nil AuditConfig is a failure as the implicit default if it is not specified
+		// should be 'disabled'
+		tr.Failf(jUnitAuditConfigClass, "There is no AuditConfig")
+		tr.Done()
```
moshloop: Done is called by the test infra - just return if no more testing is possible
```suggestion
		return
```

pkg/phases/audit/test.go
```diff
@@ -0,0 +1,144 @@
...
+	"strings"
+
+	"github.com/flanksource/commons/console"
+	"github.com/moshloop/platform-cli/pkg/platform"
+	v1 "k8s.io/api/core/v1"
+)
+
+const jUnitAuditConfigClass = "AuditConfig"
+const jUnitAuditAPIServerStateClass = "AuditApiServerState"
+const jUnitAuditLogs = "AuditLogsPresent"
```
moshloop: We can probably just group everything under a single test name
```suggestion
const testName = "auditing"
```

pkg/types/config.go
philipstaffordwood: Cool. Will do.

philipstaffordwood: AuditConfig moved into Kubernetes.
Now here:
https://github.com/flanksource/platform-cli/blob/91babc5b81de5db37b72ff5cc714a634eef2b971/pkg/types/types.go#L271-L277

pkg/phases/audit/test.go
philipstaffordwood: done

PR #245, opened 2020-04-02
fix various issues raised by linter

philipstaffordwood:
https://github.com/flanksource/platform-cli/issues/243


pkg/provision/rolling.go
```diff
@@ -84,15 +86,17 @@ func RollingUpdate(platform *platform.Platform, minAge time.Duration, drainTimeo
 
 		log.Infof("Replaced %s in %s", node.Name, timer)
 
-		doUntil(func() bool {
+		if succeededWithinTimeout := doUntil(timeout, func() bool {
```
philipstaffordwood, line 89: linter was unhappy with the `false` result (for when the check failed until timeout) was never used.

pkg/provision/vmware/clone.go
```diff
@@ -12,15 +12,15 @@ import (
 
 	cloudinit "github.com/moshloop/konfigadm/pkg/cloud-init"
 	konfigadm "github.com/moshloop/konfigadm/pkg/types"
-	. "github.com/moshloop/platform-cli/pkg/types" // nolint: golint, stylecheck
+	ptypes "github.com/moshloop/platform-cli/pkg/types"
```
philipstaffordwood, line 15: A bit ugly, happy for a better approach here. 
There is a directive here for the linter to ignore the dot import warning, but later on in `pkg/provision/vmware/vm.go` this confuses
the linter between the VM type defined there, and the VM type defined
in `pkg/types/types.go`

pkg/provision/vmware/cluster.go
```diff
@@ -58,7 +58,7 @@ func (p *vmwareCluster) GetMachinesByPrefix(prefix string) (map[string]types.Mac
 	return vms, nil
 }
 
-func (platform *vmwareCluster) GetMachine(name string) (types.Machine, error) {
-	machines, err := platform.GetMachines()
+func (p *vmwareCluster) GetMachine(name string) (types.Machine, error) {
```
philipstaffordwood, line 61: the linter wants consistent receiver names.

pkg/provision/vsphere.go
```diff
@@ -183,7 +183,7 @@ func createMaster(platform *platform.Platform) (types.Machine, error) {
 
 func createWorker(platform *platform.Platform, nodeGroup string) (types.Machine, error) {
 	if nodeGroup == "" {
-		for k, _ := range platform.Nodes {
+		for k := range platform.Nodes {
```
philipstaffordwood, line 186: linter prefers that when iterating over just the keys that the (optional) value be left out rather than _'ed

pkg/provision/vm.go
```diff
@@ -21,11 +21,11 @@ func VM(platform *platform.Platform, vm *types.VM, konfigs ...string) error {
...
 	}
 	log.Infof("Using konfigadm spec: %s\n", konfigs)
-	_vm, err := platform.Clone(*vm, konfig)
+	machine, err := platform.Clone(*vm, konfig)
 
 	if err != nil {
 		return fmt.Errorf("vm: failed to clone %v", err)
 	}
-	log.Infof("Provisioned  %s ->  %s\n", vm.Name, _vm.IP)
+	log.Infof("Provisioned  %s ->  %s\n", machine.Name(), machine.IP())
```
philipstaffordwood, line 29: switching to the new interface - so including the function braces.

https://github.com/flanksource/platform-cli/blob/3b70c36dc2ae5d56f5c213cc2fbdc456efdd543b/pkg/types/api.go#L10-L22

pkg/provision/vsphere.go
```diff
@@ -218,23 +218,24 @@ func createWorker(platform *platform.Platform, nodeGroup string) (types.Machine,
...
-			log.Warnf("[%s] failed to drain: %v", vm.Name, err)
+			log.Warnf("[%s] failed to drain: %v", vm.Name(), err)
 		}
 	}
 	client, err := platform.GetClientset()
 	if err != nil {
 		log.Warnf("Failed to get client to delete node")
 	} else {
-		client.CoreV1().Nodes().Delete(vm.Name(), &metav1.DeleteOptions{})
+		if err := client.CoreV1().Nodes().Delete(vm.Name(), &metav1.DeleteOptions{}); err != nil {
```
philipstaffordwood, line 228: linter was unhappy about an unhandled error result

PR #249, opened 2020-04-03
Added Elasticsearch dashboards and remove unmarshall to Board struct

### LINKED ISSUE #159: Grafana ELK Dashboard
* Add an elasticsearch connection with details specified in `filebeat.elasticsearch`
* Add graphs for a) Volume per indice, b) Per Cluster c) Per Namespace d) Per host

K-DOT:
Closes #159


manifests/monitoring/kube-prometheus.yaml
```diff
@@ -298,6 +298,35 @@ spec:
...
+  name: elasticsearch
+  namespace: monitoring
+spec:
+  name: elasticsearch.yaml
+  datasources:
+    - name: elasticsearch
+      type: elasticsearch
+      access: proxy
+      basicAuth: true
+      url: "https://{{.filebeat.elasticsearch.url}}:9200"
```
moshloop, line 312: I think this datasource will need to be guarded by an if statement for when filebeat/elasticsearch is not null.

The URL also shouldn't append :9200,  elastic can be server on  443/80

PR #258, opened 2020-04-07
Add consul backup/restore

teodor-pripoae:


pkg/phases/consul/tasks.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/flanksource/commons/utils"
+	"github.com/moshloop/platform-cli/pkg/k8s"
+	"github.com/moshloop/platform-cli/pkg/platform"
+)
+
+type BackupRestore struct {
+	Name      string
+	Namespace string
+	client    *k8s.Client
+	platform  *platform.Platform
```
moshloop, line 19:
```suggestion
	Name      string
	Namespace string
        *platform.Platform
```

If you make platform an embedded struct, then you can skip all the `b.client.Apply` and just use `b.Apply`

test/common.yaml
```diff
@@ -199,6 +199,9 @@ vault:
       allow_subdomains: true
       allowed_domains:
         - 127.0.0.1.nip.io
+  consul:
+    bucket: "consul-backups"
+    backupImage: "quay.io/toni0/consul-backup-restore:v0.1.2"
```
moshloop, line 204: I think default this so that you can just do:

```yaml
consul:
    bucket: consul-backups
```

PR #264, opened 2020-04-11
Add kubernetes auditing support for vSphere clusters

philipstaffordwood:
This PR extends the Kind cluster-specific auditing support introduced in https://github.com/flanksource/platform-cli/pull/233 to vSphere clusters.

The main configuration is specifying the `audit-policy-file` as an **Extra Argument** for the cluster APIServer in the ClusterConfiguration and specifying a mapping for the file for the api-server pod.

This is done in `NewClusterConfig` for both cluster types.

The injection of the audit policy file is handled differently between Kind and vSphere clusters:

For vSphere clusters these configs are injected through the `kubeadm.conf` passed to `kubeadm` 
through cloud-init images passed to the primary master using konfigadm

For Kind clusters these configs are injected as `kubeadmConfigPatches` patches specified in the Kind cluster config YAML.


pkg/provision/kind.go
```diff
@@ -133,34 +137,24 @@ func KindCluster(platform *platform.Platform) error {
...
-			MountPath: "/etc/ssl/oidc/ingress-ca.pem",
-			ReadOnly:  true,
-			PathType:  api.HostPathFile,
-		},
-	}
-	clusterConfig.APIServer.ExtraArgs["oidc-ca-file"] = "/etc/ssl/oidc/ingress-ca.pem"
 
-	if platform.Kubernetes.AuditConfig.PolicyFile != "" {
-		clusterConfig.APIServer.ExtraArgs["audit-policy-file"] = "/etc/kubernetes/policies/audit-policy.yaml"
+	vol := &clusterConfig.APIServer.ExtraVolumes
```
philipstaffordwood, line 148: since the audit configs mappings are set upstream in `NewClusterConfig` this needs to be amended to _append_ rather than _set_ the `oidc-certificate` mapping.

docs/admin-guide/audit/audit.md
```diff
@@ -150,3 +156,21 @@ nodes:
...
     ...
 </pre>
+
+### vSphere Clusters
+
+Adding the following debug keys to the platform config can grant access to the primary master
+VM to allow debugging `kubeadm init` config issues:
+
+```YAML
+debug:
```
moshloop, line 166: What was the thinking behind this? How does this differ to just using `commands:` ?

test/e2e.sh
```diff
@@ -27,7 +27,7 @@ else
 fi
 
 if [[ "$KUBECONFIG" != "$HOME/.kube/kind-config-kind" ]] ; then
-  $BIN ca generate --name ingress-ca --cert-path .certs/ingress-ca-crt.pem --private-key-path .certs/ingress-ca-key.pem --password foobar  --expiry 1
+  $BIN ca generate --name ingress-ca --cert-path .certs/root-ca.crt --private-key-path .certs/root-ca.key --password foobar  --expiry 1
```
moshloop, line 30: Why the change in filename ?

docs/admin-guide/audit/audit.md
philipstaffordwood: using the `master.commands` key is a more elegant approach. switching to it.

test/e2e.sh
philipstaffordwood: Happy to revert.
This got back-ported to `common.yaml` from the platform config I generated to get a single master cluster for audit test.

about extensions: `.key` and `.crt` seems more "standard" (if at least more common) for CAs.

about names - I was focusing on the root CA.

Ideally I should change the name from `ingress-ca` as well.

What's the verdict?
Back to `ingress-ca-crt.pem` and `ingress-ca-key.pem`,
or 
change the name to `root-ca` as well?

moshloop: +1 for the rename to .key and .crt

I think let's leave ingress-ca and add root-ca - that way we can pick any issues with using the wrong ca.

philipstaffordwood: Cool done in e9bfe90e0608705a5f75e34c5b24624674832d61

test/vsphere/vsphere-single-master.yaml
```diff
@@ -0,0 +1,52 @@
+##
```
moshloop, line 1: Not necessarily for this PR, but most of these comments should probably find their way into the godoc comments in types.go so that they pull through to https://karina.docs.flanksource.com/reference/config/

Likewise with the validations

philipstaffordwood: I've added with latest commit.

( :smile: needed an update to retrigger *test-next*)

pkg/types/config.go
```diff
@@ -1,20 +1,26 @@
...
+	// Consul specifies the endpoint for externally hosted consul cluster
+	// NOTE: a working consol config required to verify
+	//       that primary master is available for a vSphere cluster.
+	Consul         string      `yaml:"consul"`
+	Dashboard      Dashboard   `yaml:"dashboard,omitempty"`
+	Datacenter     string      `yaml:"datacenter"`
+	DNS            *DynamicDNS `yaml:"dns,omitempty"`
+	DockerRegistry string      `yaml:"dockerRegistry,omitempty"`
+	// Domain specifies the domain that cluster will be available at
+	// NOTE: domain must be supplied for vSphere clusters.
```
moshloop, line 19: It would need to be specified for kind clusters as well, otherwise, a lot of functionality will not work/deploy

pkg/phases/konfigadm.go
```diff
@@ -102,6 +104,15 @@ func addInitKubeadmConfig(platform *platform.Platform, cfg *konfigadm.Config) er
 	}
 	platform.Tracef("Using kubeadm config: \n%s", string(data))
 	cfg.Files["/etc/kubernetes/kubeadm.conf"] = string(data)
+	if platform.Kubernetes.AuditConfig.PolicyFile != "" {
+		// for vSphere clusters audit policy files are injected into the machine
+		// via the KonfigAdm
```
moshloop, line 109:
```suggestion
		// clusters audit policy files are injected into the machine via konfigadm
```

So this phase doesn't actually know what it is being used for, e.g. it could be sphere today, and AWS in the future

docs/admin-guide/audit/audit.md
```diff
@@ -150,3 +156,26 @@ nodes:
  
     ...
 </pre>
+
+### vSphere Clusters
```
moshloop, line 160: I think this section is probably more relevant in a troubleshooting doc

docs/admin-guide/audit/audit.md
```diff
@@ -34,12 +34,18 @@ The [official documentation describes](https://kubernetes.io/docs/tasks/debug-ap
 | `audit-log-maxsize`  | Specifies the maximum size in megabytes of the audit <br/>log file before it gets rotated <br/>Sets the `--audit-log-maxsize` flag.                   |
 | `audit-log-format`   | Specifies the logging format used.<br/>Options are:<br/>`"legacy"` indicates 1-line text format for each event <br/> `"json"` indicates a structured json format. <br/>Sets the `--audit-log-format` flag.                   |
 
-These are injected into the cluster at creation time via `kubeadmConfigPatches` that are supplied to `kubeadm`.
+These are injected into the cluster at creation time.
```
moshloop, line 37: I would move these entirely to source code as it is more applicable for developers, then it is for administrators who are consuming and configuring it.

philipstaffordwood: Agreed. Removed in 711b01c0d1cf5f97ab0c0b9051264b8a2b006edd

docs/admin-guide/audit/audit.md
philipstaffordwood: Agreed. Removed in 711b01c

PR #272, opened 2020-04-14
Allow a konfigadm spec to be specified as starting point for VM Configuration

philipstaffordwood:
Addresses platform-cli issue https://github.com/flanksource/platform-cli/issues/112

Incidentally fixes a minor bug where VM cloning fails where `GOVC_DATASOURCE` is not set.


pkg/provision/vmware/clone.go
```diff
@@ -88,6 +90,8 @@ func (s Session) Clone(vm ptypes.VM, config *konfigadm.Config) (*object.VirtualM
 
 	s.Infof("Cloning %s to %s", vm.Template, vm.Name)
 
+	log.Tracef("VM Spec: %# v", pretty.Formatter(spec))
```
moshloop, line 93: Use s.Tracef and not logrus directly

pkg/phases/konfigadm.go
```diff
@@ -56,10 +56,21 @@ func CreatePrimaryMaster(platform *platform.Platform) (*konfigadm.Config, error)
 
 func baseKonfig(platform *platform.Platform) (*konfigadm.Config, error) {
 	platform.Init()
-	cfg, err := konfigadm.NewConfig().Build()
-	if err != nil {
-		return nil, fmt.Errorf("baseKonfig: failed to get config: %v", err)
+	var cfg *konfigadm.Config
+	var err error
+
+	if platform.Master.KonfigadmFile == "" {
```
moshloop, line 62: This would use the master konfigadm file for all VM's? If I specify a konfigadm file under master, I wouldn't expect it to be users under workers and vice versa. You may also want to use different konfigadm for different worker pools

pkg/provision/vmware/clone.go
philipstaffordwood: :+1:

pkg/phases/konfigadm.go
philipstaffordwood: :thinking: , yes.

Cool. Will rework this into being specific per master or worker pool.

pkg/phases/konfigadm.go
```diff
@@ -40,37 +46,110 @@ func CreatePrimaryMaster(platform *platform.Platform) (*konfigadm.Config, error)
...
 	// update hosts file with hostname
-	cfg.AddCommand("echo $(ifconfig ens160 | grep inet | awk '{print $2}' | head -n1 ) $(hostname) >> /etc/hosts")
+	cfg.AddCommand(updateHostsFileCmd)
 	return cfg, nil
 }
 
+// addAuditConfig derives the initial admin config for a cluster from its platform
+// config and adds it to its konfigadm files
+func addAuditConfig(platform *platform.Platform, cfg *konfigadm.Config) error {
+	if platform.Kubernetes.AuditConfig.PolicyFile != "" {
```
moshloop, line 140:
```suggestion
	if platform.Kubernetes.AuditConfig != nil && platform.Kubernetes.AuditConfig.PolicyFile != "" {
```

test/linter.yaml
```diff
@@ -5,3 +5,9 @@ yamlCheck:
     - mkdocs.yml
     - .git/(.*)
     - .golangci.yml
+# Exclude konfigadm-template.yml from linting
```
moshloop, line 8: https://github.com/flanksource/konfigadm/pull/62 has merged

pkg/types/types.go
```diff
@@ -50,7 +51,10 @@ type VM struct {
 	// Tags to be applied to the VM
 	Tags     map[string]string `yaml:"tags,omitempty"`
 	Commands []string          `yaml:"commands,omitempty"`
-	IP       string            `yaml:"-"`
+	// KonfigadmFile allows for the specification of a konfigadm
+	// spec used as basis for configuring the VM on creation.
```
moshloop, line 55:
```suggestion
	// A path to a konfigadm specification used for configuring the VM on creation.
```

`{Fieldname} comment` is go convention to make godocs readable at a summary level, but the docs we create from these comments are in a table with the Fieldname as the first column so adding it is suplerfous

PR #327, opened 2020-05-18
Add opa labels to all platform namespaces

teodor-pripoae:


moshloop:
Maybe we should change the CreateOrUpdateNamespace to:

```golang
if annotations == nil {
annotations = DefaultPlatformAnnotations()
}
```

moshloop:
@teodor-pripoae Can we move the gitops test repo to flanksource/gitops-test when you update the annotations?

Issue #247, opened 2020-04-02
[E2E] Minimal configuration


moshloop:
Create an E2E that will pick up issues with empty configuration (normally nil pointers)

philipstaffordwood:
work happening in this [fork and branch](https://github.com/flanksource/platform-cli/compare/master...philipstaffordwood:issue-247-minimal-e2e)

PR #342, opened 2020-05-27
Minimal e2e for Kind

philipstaffordwood:
Adds a GitHub Actions test for spinning up a KIND cluster with minimal configs to be able to flush out config bugs.

Takes some error handling changes developed for the vSphere e2e tests in https://github.com/flanksource/karina/pull/287

Partly covers Issue #247


.github/workflows/kind.yml
```diff
@@ -0,0 +1,26 @@
+name: Kind e2e Tests
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  test-e2e:
+    runs-on: ubuntu-latest
```
philipstaffordwood, line 9: @moshloop I switched back to the VM. 
Seem simpler/faster for now to use the VM - which is basically with [everything and the kitchen sink](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md) already installed - than to get Kind working in a docker setup *and* updating build tools to the latest git and pushing to docker.

.github/workflows/kind.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - name: Run e2e testing script
+        env:
+          GIT_API_KEY: ${{ secrets.GIT_API_KEY }}
```
philipstaffordwood, line 14: @moshloop we need to add the GIT_API_KEY to the repo secrets for GH Actions

philipstaffordwood: I'm keen to make the PR Comments a bit less noisy using [summaries](https://github.com/flanksource/build-tools/issues/9#issuecomment-634554506) and to make the reporting a bit more [obvious](https://github.com/flanksource/build-tools/pull/8) in terms of status, when the E2E vSphere testing is done.

PR #345, opened 2020-05-29
Journalbeat, Auditbeat, EventRouter, Packetbeat

teodor-pripoae:


pkg/types/types.go
```diff
@@ -474,15 +474,40 @@ type FluentdOperator struct {
 }
 
 type Filebeat struct {
+	Enabled
```
moshloop, line 477: Can we switch these all to Disabled?

i.e. if a Version is specified, then it is enabled, to disable it either remove the config, or specify:    
`-e filebeat.disabled=true`

PR #412, opened 2020-07-06
switch to fixed cluster name

philipstaffordwood:


moshloop:
This is looking better - Can we just ensure that the check fails if any of tests fail - I see minio is failing (correctly)

PR #418, opened 2020-07-10
configure virtual serial device for vsphere VMs

### LINKED ISSUE #336: Add or configure serial device on vSphere VMs
Ubuntu based VMs in vSphere need a serial device to boot properly.

Currently this needs to be manually configured on a konfigadm-supplied VM templates.

Adding a short section adding/configuring the serial device (similar to how the cdrom device is handled) would make this a bit more bulletproof.

philipstaffordwood:
closes #336


moshloop:
Can we use a network based serial port that doesn't require a backing file?

Screenshot 2020-07-10 at 16 43 13

philipstaffordwood:
> Can we use a network based serial port that doesn't require a backing file?
> 
> Screenshot 2020-07-10 at 16 43 13

I think we could.
Where do we point it to?

philipstaffordwood:
@moshloop, I've implemented the network serial port config similar to the screenshot.

It works - i.e. provisioning succeeds.
I tried to deploy two clusters with the same config - also works.

I'm not sure how the implementation works in the back-end, so I'm not sure how it will perform/behave with nothing actually on `localhost:0`, but a simple spin up and see if anything breaks test succeeds.

PR #440, opened 2020-07-16
Fix k8s dashboard

teodor-pripoae:


manifests/k8s-dashboard.yaml
```diff
@@ -171,8 +171,8 @@ metadata:
     kubernetes.io/tls-acme: "true"
     nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
-    {{ if .oauth2Proxy.enabled }}
-    nginx.ingress.kubernetes.io/auth-url: "https://oauth2.{{.domain}}/oauth2/auth"
+    {{ if not .oauth2Proxy.disabled }}
+    nginx.ingress.kubernetes.io/auth-url: "http://oauth2-proxy.ingress-nginx.svc.cluster.local:4180/oauth2/auth"
```
moshloop, line 175: Shouldn't this be https?

teodor-pripoae: No, I changed it to use internal svc IP due to issues with certificates. The `auth-url` is called internally by nginx, then the user is redirected to`auth-signin` url.

PR #450, opened 2020-07-28
multi cluster support for kube resource report and kube-web-view

### LINKED ISSUE #431: Kube Resource Report - Multi Cluster Support
- Generate a new cert and then map it to an appropriate read-only role
- Create the kubeconfig  for each specified cluster

### LINKED ISSUE #432: Kube Web View - Multi Cluster Support

### LINKED ISSUE #433: Kube Web/Resource View - Read cluster name
See https://github.com/flanksource/canary-checker/blob/master/pkg/kube.go#L46

### LINKED ISSUE #452: [bug] kube-web-view and kube-resource-report give SSL errors connecting to own cluster with service account
If no kubeconfig file is injected both kube-web-view and kube-resource-report fail to connect to the K8S API with an SSL error:

image

image

philipstaffordwood:
Adds multi-cluster support for kube-resource-report and kube-web-view by injecting kubeconfig files to specify connectivity

- Closes #431 
- Closes #432

Incidentally allows overriding of cluster names:
- Closes #433

And resolves cluster API endpoint issues in these when using the default in-cluster service account:
- Closes #452


moshloop:
I think the ClusterRole can be static i.e. it gets created on every cluster that kube-resource-report is deployed to, then only the kubeconfig is created and deleted when extraClusters are specified

philipstaffordwood:
> I think the ClusterRole can be static i.e. it gets created on every cluster that kube-resource-report is deployed to, then only the kubeconfig is created and deleted when extraClusters are specified

Cool. So it will be deployed on each cluster that it reports on.

philipstaffordwood:
I'm including https://github.com/flanksource/karina/issues/433 in this PR since the simplest way to specify the cluster name is by using a kubeconfig file.
So commit https://github.com/flanksource/karina/pull/450/commits/137a5cf9f34e96e02a6e69b1f17e18bc3df5b4fc just removes the conditional logic to NOT use a kubeconfig file and aligns the context name to the cluster name in the generated kubeconfig

pkg/k8s/client.go
```diff
@@ -1699,3 +1746,19 @@ func (c *Client) GetHealth() Health {
...
 }
+
+// GetExternalClient constructs a Client for accessing an external cluster.
+// It uses the given CA, clustername and cluster host of the cluster API endpoint
+// to configure connectivity.
+func GetExternalClient(logger logger.Logger, clusterName string, clusterHost string, cacert *certs.Certificate) *Client { //nolint interfacer
+	return &Client{
+		Logger: logger,
+		GetKubeConfigBytes: func() ([]byte, error) {
+			kubeConfig, err := CreateKubeConfig(clusterName, cacert, clusterHost, "system:masters", "admin", 24*7*time.Hour)
```
moshloop, line 1757: What happens after 7 days ?

philipstaffordwood: ~~This one is cool - it doesn't get written to disk and is ephemeral. Only used in this one session to connect to external cluster.~~
This now unused - since we no longer create remote RBAC,

And this does apply to the kubeconfig below.

pkg/phases/kuberesourcereport/install.go
```diff
@@ -3,25 +3,69 @@ package kuberesourcereport
...
+	ca, err := ca.ReadCA(p.CA)
+	if err != nil {
+		return fmt.Errorf("unable to get root CA %v", err)
+	}
+	// kube-resource-view can't use the service account to access it's own cluster
+	// so we add user/cert access via the default internal API endpoint
+	p.KubeResourceReport.ExternalClusters.AddSelf(p.Name)
+	// create a secret containing a kubeconfig file that allows access to
+	// this cluster via user/cert as well as the given external clusters
+	kubeConfig, err := k8s.CreateMultiKubeConfig(ca, p.KubeResourceReport.ExternalClusters, Group, User, 24*7*time.Hour)
```
philipstaffordwood, line 56: what expiry would be acceptable here?

pkg/phases/kubewebview/install.go
```diff
@@ -2,26 +2,69 @@ package kubewebview
...
+	ca, err := ca.ReadCA(p.CA)
+	if err != nil {
+		return fmt.Errorf("unable to get root CA %v", err)
+	}
+	// kube-web-view can't use the service account to access it's own cluster
+	// so we add user/cert access via the default internal API endpoint
+	p.KubeWebView.ExternalClusters.AddSelf(p.Name)
+	// create a secret containing a kubeconfig file that allows access to
+	// this cluster via user/cert as well as the given external clusters
+	kubeConfig, err := k8s.CreateMultiKubeConfig(ca, p.KubeWebView.ExternalClusters, Group, User, 24*7*time.Hour)
```
philipstaffordwood, line 55: what expiry would be acceptable here?

PR #506, opened 2020-10-12
add self-hosted runners

moshloop:

PR #508, opened 2020-10-13
Kind cache configure

BrendanGalloway:
@moshloop Once this merges you should be able to update #506 to export CACHE_DIR=/runner/_work in order for the docker container to see them.


moshloop:
I don't think there are any go.mod/sum changes necessary for this PR - can we remove them ?

BrendanGalloway:
> I don't think there are any go.mod/sum changes necessary for this PR - can we remove them ?

Done

PR #525, opened 2020-10-23
Wait for reconfig-map controller to start

BrendanGalloway:


pkg/phases/configmapreloader/install.go
```diff
@@ -24,5 +25,12 @@ func Deploy(p *platform.Platform) error {
 	if err := p.CreateOrUpdateNamespace(constants.PlatformSystem, nil, nil); err != nil {
 		return err
 	}
-	return p.ApplySpecs("", "configmap-reloader.yaml")
+
+	if err := p.ApplySpecs("", "configmap-reloader.yaml"); err != nil {
+		return err
+	}
+
+	p.WaitForNamespace(constants.PlatformSystem, 60*time.Second)
```
moshloop, line 33: This might introduce some unintended delays if there are other resources in platform-system not healthy during a deployment that doesn't actually change configmap-reloader - The easiest fix would be to move the wait into the test vs the deployment

BrendanGalloway:
squashed into #531

PR #531, opened 2020-10-26
Various fixes to correctly wait for systems to start during testing

BrendanGalloway:


moshloop:
I think maybe bump the timeouts to 3m

PR #572, opened 2020-12-07
Add volumesnapshotlocation config to velero manifest

BrendanGalloway:


manifests/velero.yaml
```diff
@@ -50,6 +50,10 @@ metadata:
   namespace: velero
 spec:
   provider: aws
+  {{ if .velero.config.region }}
```
moshloop, line 53: I think make this a full copy of the config e.g.

{{ range $k, $v := .velero.config }} 
$k: "$v"
{{end }}

PR #599, opened 2021-01-18
Add istio deployment

paddatrapper:
Fixes: #585


manifests/istio-operator.yaml
```diff
@@ -0,0 +1,257 @@
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: istio-operator
+  name: istio-operator
+---
+# SYNC WITH manifests/charts/base/files
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
moshloop, line 214: Can we move the CRD definition out and put it into manifest/crds folder and update the crd phase ?

pkg/types/default.go
```diff
@@ -9,6 +9,9 @@ func DefaultPlatformConfig() PlatformConfig {
 		CertManager: CertManager{
 			Version: "v1.0.3",
 		},
+		IstioOperator: IstioOperator{
```
moshloop, line 12: I don't think this should be installed by default

paddatrapper: Any suggestions about which test to use?

pkg/phases/istiooperator/install.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/flanksource/karina/pkg/platform"
+)
+
+const (
+	Namespace = "istio-operator"
+)
+
+func Install(p *platform.Platform) error {
```
moshloop, line 11: I think these need to be added to both cmd.Deploy and cmd.Test

manifests/istio-operator.yaml
```diff
@@ -0,0 +1,211 @@
...
+      name: istio-operator
+  template:
+    metadata:
+      labels:
+        name: istio-operator
+    spec:
+      serviceAccountName: istio-operator
+      containers:
+        - name: istio-operator
+          image: docker.io/istio/operator:1.8.2
```
moshloop, line 144:
```suggestion
          image: docker.io/istio/operator:{{.istioOperator.version | default "na" }}
```

pkg/types/default.go
paddatrapper: Use platform

paddatrapper:
Looks like Istio is installed and passes the test in the E2E/test for platform

PR #616, opened 2021-01-26
ci: Accurate reporting of vsphere test status

paddatrapper:


paddatrapper:
The test still passes even though karina fails and the test exits 1, but it is now grouped and the ssh logic is fixed

PR #634, opened 2021-01-31
Add kibana headers for sso

teodor-pripoae:


manifests/elasticsearch.yaml
```diff
@@ -90,6 +90,12 @@ metadata:
   annotations:
     nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
     kubernetes.io/tls-acme: "true"
+    {{ if not (index .oauth2Proxy "disabled") }}
+    platform.flanksource.com/restrict-to-groups: {{ join .oauth2Proxy.oidcGroups ";" }}
```
moshloop, line 94: I think lets add a []string UserGroups under Ldap and then use Ldap.AdminGroup and UserGroups

moshloop (changes requested):

PR #652, opened 2021-02-06
Downgrade default dex version to fix image pull error

teodor-pripoae:


moshloop:
I think lets rather switch the image location to ghcr.io/dexidp/dex

PR #653, opened 2021-02-07
feat: add ArgoCD Operator Production runtime

### LINKED ISSUE #639: Add Argo deployment

tobernguyen:
**Describe**
- Add ArgoCD Operator as a Production Runtime

**How to test**:
To install ArgoCD Operator, specify the version of ArgoCD Operator in your `cluster-config.yml`:
```yaml
domain: 127.0.0.1.nip.io
name: test-cluster
argoCDOperator:
  version: v0.0.15
calico:
  version: v3.8.2
kubernetes:
  version: v1.16.9
  masterIP: localhost
  containerRuntime: containerd
ca:
  cert: .certs/root-ca.crt
  privateKey: .certs/root-ca.key
  password: foobar
ingressCA:
  cert: .certs/ingress-ca.crt
  privateKey: .certs/ingress-ca.key
  password: foobar
```

**Deploy ArgoCD Operator with**
```zsh
karina deploy phases --crds --argocd-operator -c cluster-config.yml
```

**Test ArgoCD Operator Production Runtime with**
```zsh
karina test phases --e2e --argocd-operator -c cluster-config.yml
```

This test phase will deploy a minimal ArgoCD cluster then check for the ArgoCD cluster health by actively looking for all expected components of ArgoCD Cluster to come up healthy (4 Deployments including application-controller, dex-server, redis, repo-server, and server).

Closes #639


pkg/types/config.go
```diff
@@ -3,14 +3,15 @@ package types
...
 type PlatformConfig struct {
-	Brand         Brand          `yaml:"brand,omitempty" json:"brand,omitempty"`
-	Version       string         `yaml:"version" json:"version,omitempty"`
-	Velero        Velero         `yaml:"velero,omitempty" json:"velero,omitempty"`
-	CA            *CA            `yaml:"ca" json:"ca,omitempty"`
-	CanaryChecker *CanaryChecker `yaml:"canaryChecker,omitempty" json:"canaryChecker,omitempty"`
-	Calico        *Calico        `yaml:"calico,omitempty" json:"calico,omitempty"`
-	Antrea        *Antrea        `yaml:"antrea,omitempty" json:"antrea,omitempty"`
-	CertManager   CertManager    `yaml:"certmanager,omitempty" json:"certmanager,omitempty"`
+	ArgoCDOperator ArgoCDOperator `yaml:"argoCDOperator,omitempty" json:"argoCDOperator,omitempty"`
```
moshloop, line 6:
```suggestion
	ArgocdOperator ArgocdOperator `yaml:"argocdOperator,omitempty" json:"argocdOperator,omitempty"`
```

pkg/phases/argocdoperator/test.go
```diff
@@ -0,0 +1,88 @@
...
+	defer removeE2EArgoCluster(p, clusterName, test)
+	if err != nil {
+		test.Failf(testName, "Error creating ArgoCD Cluster %s: %v", clusterName, err)
+		return
+	}
+	test.Passf(testName, "Cluster %s deployed", clusterName)
+
+	test.Infof("Checking if ArgoCD cluster is healthy...")
+	// List of expected deployment to be deployed by ArgoCD Operator for ArgoCD Cluster
+	expectedDeployments := []string{
```
moshloop, line 55: I think it might be simpler to move the fmt.Sprintf into the loop

pkg/phases/argocdoperator/test.go
```diff
@@ -0,0 +1,88 @@
...
+	client, _ := p.GetClientset()
+	kommons.TestNamespace(client, Namespace, test)
+	if p.E2E {
+		TestE2E(p, test)
+	}
+}
+
+func TestE2E(p *platform.Platform, test *console.TestResults) {
+	testName := "argocd-operator-e2e"
+	if p.ArgoCDOperator.IsDisabled() {
```
moshloop, line 39: I don't think this will ever be true ?

pkg/phases/argocdoperator/install.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func Deploy(platform *platform.Platform) error {
+	if platform.ArgoCDOperator.IsDisabled() {
+		if err := platform.DeleteSpecs("", "argocd-operator.yaml"); err != nil {
+			platform.Warnf("failed to delete specs: %v", err)
+		}
+		return nil
+	}
+	if platform.ArgoCDOperator.Version == "" {
```
moshloop, line 18: This will never be true, a component without a version is considered disabled. Defaulting would need be to done above the IsDisabled check() or in defaults.go

Argo wouldn't be a critical component so it would be disabled by default

pkg/phases/argocdoperator/install.go
```diff
@@ -0,0 +1,27 @@
...
+const (
+	Namespace = "argocd"
+)
+
+func Deploy(platform *platform.Platform) error {
+	if platform.ArgoCDOperator.IsDisabled() {
+		if err := platform.DeleteSpecs("", "argocd-operator.yaml"); err != nil {
+			platform.Warnf("failed to delete specs: %v", err)
+		}
+		return nil
```
moshloop, line 16: Swallowing deployment failures should only be by exception, not by default
```suggestion
		return platform.DeleteSpecs("", "argocd-operator.yaml"); 
```

manifests/argocd-operator.yaml
```diff
@@ -0,0 +1,3196 @@
+# This yaml is built by this command: kustomize build "https://github.com/argoproj-labs/argocd-operator/deploy?ref=v0.0.14"
```
moshloop, line 1: Can we split the CRDS out into the crd phase?

pkg/api/argocd/argocd_types.go
```diff
@@ -0,0 +1,481 @@
...
+	Image string `json:"image,omitempty"`
+
+	// Ingress defines the desired state for an Ingress for the Grafana component.
+	Ingress ArgoCDIngressSpec `json:"ingress,omitempty"`
+
+	// Resources defines the Compute Resources required by the container for Grafana.
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Route defines the desired state for an OpenShift Route for the Grafana component.
+	Route ArgoCDRouteSpec `json:"route,omitempty"`
```
moshloop, line 125: I don't think we want to add a dependency on OpenShift API, so we can remove this field.
If a route really is required then a new template from the template operator can be used to back k8s ingress with routes

manifests/argocd-operator.yaml
```diff
@@ -0,0 +1,3196 @@
...
+  name: argocd-server
+subjects:
+- kind: ServiceAccount
+  name: argocd-server
+  namespace: argocd
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: argocd-operator
```
moshloop, line 3167: This should be namespaced into `platform-system`

manifests/argocd-operator.yaml
```diff
@@ -0,0 +1,3196 @@
...
+          type: object
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+---
+apiVersion: v1
+kind: ServiceAccount
```
moshloop, line 2746: Are these argocd rbac objects not created dynamically by the operator based on the namespace the ArgoCD CRD is found in?

pkg/phases/argocdoperator/install.go
moshloop: The version will need to be added to the tests/cicd.yaml file to enable tests for that suite.

moshloop (changes requested):

pkg/phases/argocdoperator/install.go
tobernguyen: I know doing this will propagate the exception to the function calling this Deploy function but I'm not sure I fully understand what you are saying. Can you explain more or give an example?

tobernguyen: We are having this same logic in postgresoperator/install.go, should we refactor that to make it consistent?

pkg/phases/argocdoperator/install.go
tobernguyen: > This will never be true, a component without a version is considered disabled. Defaulting would need be to done above the IsDisabled check() or in defaults.go
> 
> Argo wouldn't be a critical component so it would be disabled by default

Correct this will never be true. I just use the same pattern from `postgresloperator/install.go` 😂 . After a while I realized this but I forgot to fix. Thanks for catching this.

pkg/phases/argocdoperator/test.go
tobernguyen: So if the user defines `argocdOperator.disabled: true` or they didn't define `argocdOperator.version` then this will be true.

pkg/api/argocd/argocd_types.go
tobernguyen: Ok, so I will remove this Route to remove `openshift` dependency. Do we want to make a note somewhere about things we modified from upstream. Doing so will help with future updates from upstream.

manifests/argocd-operator.yaml
tobernguyen: I can do that.

pkg/phases/argocdoperator/test.go
moshloop: correct

pkg/phases/argocdoperator/install.go
moshloop: Yes - that makes sense

pkg/api/argocd/argocd_types.go
moshloop: That's a good idea, maybe in the form of a patch?

tobernguyen: I'm not sure how to patch this. But I'm commenting out the code and write a note about the decision. At least the next time we update from upstream, the comment will be in the git diff so we can be aware of the change. If you have a better idea please let me know.

manifests/argocd-operator.yaml
tobernguyen: Right now the `postgres-operator` is being installed into `postgres-operator` namespace. Same for redisoperator, rabbitmqoperator. So looks like each operator will have its own namespace. Do you really want to move argocd-operator to `platform-system` which doesn't follow the current convention?

manifests/argocd-operator.yaml
tobernguyen: This is a really good catch. Looks like they did that in the latest tag (v0.0.14): https://github.com/argoproj-labs/argocd-operator/blob/v0.0.14/deploy/role.yaml but they fixed this (remove all rbac objects of ArgoCD Cluster from the operator) in master branch: https://github.com/argoproj-labs/argocd-operator/blob/master/deploy/role.yaml

I will generate this file again from `master`. Then will separate CRDs to CRDs phase as well.

tobernguyen: Looks like removing the RBAC object break the funationality of the operator + argocd cluster. Seems like a bug from upstream. I will dig into this tomorrow.

Ref: https://github.com/argoproj-labs/argocd-operator/pull/224

tobernguyen: So basically, they remove that ArgoCD Cluster's RBCA objects from manifest and make the operator create them in the above PR but they haven't released a version with that change yet. We might need to wait for them to release a new version that includes that fix.

moshloop: I think lets at least split the RBAC for the operator vs the RBAC for the clusters that come up into separate files

tobernguyen: I have removed the RBAC of the cluster from installing phase and split them into a separate file and in the test phase, I use that file to deploy necessary RBAC for the cluster it can come up healthy.

manifests/argocd-operator.yaml
tobernguyen: I will keep the namespace of ArgoCD Operator to `argocd` for now as it follows the current convention of other operators. Let me know if you want to do something differently.

PR #659, opened 2021-02-08
Make test name unique

paddatrapper:
For the tests not covered by https://github.com/flanksource/kommons/pull/14


paddatrapper:
Need to fix kube-dns, kube-state-metrics, kubelet

paddatrapper:
That should be all the duplicates removed. If there are any still around, we will be able to notice them when the results fail to be uploaded to tesults

Issue #458, opened 2020-08-04
karina cleanup-jobs 


moshloop:
Remove all failed jobs in a given namespace

PR #682, opened 2021-02-11
feat: Remove all failed jobs in a given namespace

git-hyagi:
Proposal solution for issue #458.

### Description

Creates a new command (expecting the `--namespace` as a flag) to remove all jobs with a state not equal to "_succeeded_" on the provided namespace.

### Breaking Change

- [ ] Yes
- [X] I believe not (but, didn't have a chance to test)

### Testing Notes

I didn't test it through the **karina** commands, but I did some tests of the `BatchV1().Jobs` in an **OpenShift/Kubernetes** environment.

### **Links**

https://github.com/flanksource/karina/issues/458

### DIFF
--- a/cmd/cleanup-jobs.go
+++ b/cmd/cleanup-jobs.go
@@ -1,0 +1,64 @@
+package cmd
+
+import (
+	"context"
+	"sync"
+
+	"github.com/spf13/cobra"
+	v1 "k8s.io/api/batch/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// CleanupJobs removes all failed jobs in a given namespace
+var CleanupJobs = &cobra.Command{
+	Use:   "cleanupjobs",
+	Short: "remove all failed jobs in a given namespace",
+	Run: func(cmd *cobra.Command, args []string) {
+
+		namespace, _ := cmd.Flags().GetString("namespace")
+
+		p := getPlatform(cmd)
+		clientSet, err := p.GetClientset()
+		if err != nil {
+			p.Fatalf("Failed to create the new k8s client: %v", err)
+		}
+
+		// gather the list of jobs from a namespace
+		jobs, err := clientSet.BatchV1().Jobs(namespace).List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			p.Fatalf("Failed to gather the list of jobs from namespace %v: %v", namespace, err)
+		}
+
+		// WaitGroup to synchronize go routines execution
+		wg := sync.WaitGroup{}
+
+		// loop through the list of jobs found
+		for _, j := range jobs.Items {
+			wg.Add(1)
+
+			go func(j v1.Job, clientSet *kubernetes.Clientset) {
+
+				// loop through the job object's status conditions
+				for _, conditions := range j.Status.Conditions {
+
+					// if the type of the JobCondition is equal to "Failed", delete the job
+					if conditions.Type == "Failed" {
+						p.Infof("Removing failed job %v from namespace %v. Failed reason: %v", j.Name, j.Namespace, conditions.Reason)
+						if err = clientSet.BatchV1().Jobs(namespace).Delete(context.TODO(), j.Name, metav1.DeleteOptions{}); err != nil {
+							p.Errorf("Failed to delete job: %v", err)
+						}
+						break
+					}
+				}
+				wg.Done()
+			}(j, clientSet)
+		}
+
+		wg.Wait()
+	},
+}
+
+func init() {
+	CleanupJobs.Flags().String("namespace", "", "Namespace to cleanup failed jobs.")
+}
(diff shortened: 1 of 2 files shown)


cmd/cleanup-jobs.go
```diff
@@ -0,0 +1,62 @@
...
+		wg := sync.WaitGroup{}
+
+		// loop through the list of jobs found
+		for _, j := range jobs.Items {
+			wg.Add(1)
+
+			go func(j v1.Job, clientSet *kubernetes.Clientset) {
+
+				// if job status is not equal to "succeeded", delete the job
+				if j.Status.Failed == 0 {
```
BrendanGalloway, line 40: This check seems wrong - if the job failed, then the count of failed pods would be > 0?  I would also recommend checking though JobConditions or similar that the job has completed before attempting to delete

cmd/cleanup-jobs.go
```diff
@@ -0,0 +1,62 @@
...
+					log.Printf("INFO: Removing job %v from namespace %v", j.Name, j.Namespace)
```
BrendanGalloway, line 41: Please use the log wrapper in the platform struct with the appropriate wrapper eg:
p.Debugf() or p.Infof()

cmd/cleanup-jobs.go
```diff
@@ -0,0 +1,62 @@
...
+
+		wg.Wait()
+	},
+}
+
+func init() {
+	CleanupJobs.Flags().String("namespace", "", "Namespace to cleanup failed jobs.")
+}
+
+// function to print error message and abort program execution (os.Exit(1))
```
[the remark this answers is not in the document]
git-hyagi: Thank you for reviewing the code! :smile:

The `exit(1)` was not meant to be returned by the function. From the `log.Fatalf` doc:
> "Fatalf is equivalent to Printf() followed by a call to os.Exit(1)." 

Anyways, I'm removing this function to avoid confusions/mistakes and will follow the same pattern as from the project:
~~~
if err:=.... ; err != nil {
~~~

cmd/cleanup-jobs.go
git-hyagi: Got it! My bad, sorry for not following that.

cmd/cleanup-jobs.go
git-hyagi: You're right! 👍
I was doing some tests and forgot to change it back.

cmd/cleanup-jobs.go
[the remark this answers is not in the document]
BrendanGalloway: I was not aware of that - I have always explicitly exited 1.  In that case, the check() function would not be appropriate for use on line 43.  Failing to delete one of a list of jobs should not be fatal

cmd/cleanup-jobs.go
```diff
@@ -36,11 +40,17 @@ var CleanupJobs = &cobra.Command{
...
-					err = clientSet.BatchV1().Jobs(namespace).Delete(context.TODO(), j.Name, metav1.DeleteOptions{})
-					check(err)
+				// loop through the job object's status conditions
+				for _, conditions := range j.Status.Conditions {
+
+					// if the type of the JobCondition is equal to "Failed", delete the job
+					if conditions.Type == "Failed" {
+						log.Infof("Removing failed job %v from namespace %v. Failed reason: %v", j.Name, j.Namespace, conditions.Reason)
+						if err = clientSet.BatchV1().Jobs(namespace).Delete(context.TODO(), j.Name, metav1.DeleteOptions{}); err != nil {
+							log.Fatalf("Failed to delete job: %v", err)
```
BrendanGalloway, line 50: This should be Errorf() rather than Fatalf - rather attempt to delete all failed jobs and report which could or could not be deleted than hard exit in the middle of a concurrent loop.

BrendanGalloway, line 50: Please use the log wrapper in the platform struct

BrendanGalloway (changes requested):

BrendanGalloway:
@git-hyagi thanks for the changes - @moshloop will do the final review and merge, but in the interim can I ask you to please rebase this onto master and squash into a single `feat:` commit.

git-hyagi:
> "[..] can I ask you to please rebase this onto master and squash into a single `feat:` commit."

Done.

moshloop:
@git-hyagi please can you ping me on  slack ?

cmd/cleanup-jobs.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/spf13/cobra"
+	v1 "k8s.io/api/batch/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// CleanupJobs removes all failed jobs in a given namespace
+var CleanupJobs = &cobra.Command{
+	Use:   "cleanupjobs",
```
moshloop, line 15: Can we move this to `karina cleanup jobs` - It does conflict with `karina cleanup` which we should probably rename to `karina terminate-cluster`

PR #686, opened 2021-02-12
Add Argo Rollouts

### LINKED ISSUE #655: Deploy Argo Rollouts

tobernguyen:
### Description

- Add Argo Rollouts as a Production Runtime
- Add a make command (under `manifests/Makefile`) to generate manifest files based off the yaml files in `upstream` directory.
  ```
  cd manifests
  make
  ```

### Dependencies

- None

### Breaking Change

- [x] No

### Testing Notes

**What I did:**
- Create a `cluster-config.yml` and specify a version of Argo Rollouts
```yaml
domain: 127.0.0.1.nip.io
name: test-cluster
calico:
  version: v3.8.2
kubernetes:
  version: v1.18.15
  masterIP: localhost
  containerRuntime: containerd
ca:
  cert: .certs/root-ca.crt
  privateKey: .certs/root-ca.key
  password: foobar
ingressCA:
  cert: .certs/ingress-ca.crt
  privateKey: .certs/ingress-ca.key
  password: foobar
argoRollouts:
  version: v0.10.2
```
- Deploy Argo Rollouts with
```
karina deploy phases --crds --argo-rollouts -c cluster-config.yml
```
- Test Argo Rollouts Production Runtime with
```
karina test phases --e2e --argo-rollouts -c cluster-config.yml
```

Same

### **Links**

Closes #655


manifests/argo-rollouts.yaml
```diff
@@ -0,0 +1,293 @@
+# This yaml was generated from this command: kustomize build "https://github.com/argoproj/argo-rollouts/manifests/cluster-install?ref=v0.10.2"
```
moshloop, line 1: Make move this to a manifests/Makefile command: `make argo` ?

manifests/argo-rollouts.yaml
```diff
@@ -0,0 +1,293 @@
...
+    namespace: argo-rollouts
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/component: server
+    app.kubernetes.io/name: argo-rollouts-metrics
+    app.kubernetes.io/part-of: argo-rollouts
+  name: argo-rollouts-metrics
```
moshloop, line 258: The service and deployment should be deployed into platform-system

tobernguyen: We are having an opening conversation about this here: https://github.com/flanksource/karina/pull/653#discussion_r571559557. But this one is a bit different, it is not technically an "operator". Isolate workload to its own namespace is a good practice too. But I'm not really sure either. What do you think?

moshloop: The current model we are using to try and consolidate "platform" related workloads into platform-system to reduce namespace sprawl as it can impact network resources for those CNI's that allocate routers/subnets per NS.

i.e. single controllers/deployments don't deserve their own namespace

We do have https://github.com/flanksource/karina/issues/87 - it probably makes sense to create a test that fails on namespaces created after a deployment that is not whitelisted in an architecture document where their isolation from platform-system can be justified and documented.

tobernguyen: That makes sense. I will make a change to deploy argorollouts to `platform-system`. But I think I will change the install namespace in the `install.go` (https://github.com/flanksource/karina/pull/686/files#diff-e26e2b56e5b4e1266b6675476562c7888ac0bed17043bd17895f0d5780193532R8) file instead of directly modifying yaml files.

manifests/argo-rollouts.yaml
tobernguyen: I can try to do that. But that also means in that make command we will need to build 2 yaml files:
- `manifests/argo-rollouts.yaml` for all resources of argorollouts except CRDs
- `manifests/crd/argo-rollouts.yaml` for CRDs of argorollouts

Is that correct?

moshloop: Yeh - If make/yq isn't up to the job it could be done in a hack/update.go as well

tobernguyen: Implemented

manifests/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+define ARGO_ROLLOUTS_INSTALL_KUSTOMIZATION
+bases:
+  - github.com/argoproj/argo-rollouts/manifests/base?ref=$(ARGO_ROLLOUTS_VERSION)
+  - github.com/argoproj/argo-rollouts/manifests/role?ref=$(ARGO_ROLLOUTS_VERSION)
+resources:
+  - https://raw.githubusercontent.com/argoproj/argo-rollouts/$(ARGO_ROLLOUTS_VERSION)/manifests/cluster-install/argo-rollouts-clusterrolebinding.yaml
+images:
+  - name: argoproj/argo-rollouts
+    newTag: "{{.argoRollouts.version}}"
```
moshloop, line 20: Where does this come from ?

tobernguyen: This will result in this line: https://github.com/flanksource/karina/pull/686/files#diff-a11ce6911899b023c2d194687227963b52ebc30e81c5916471f2342d52291800R286 then it will be populated from the `cluster-config.yml` during the install phase.
```
domain: 127.0.0.1.nip.io
name: test-cluster
....
argoRollouts:
  version: v0.10.2
```

moshloop: ah makes sense. 
We probably also want to rename  argoproj/argo-rollouts to docker.io/argoproj/argo-rollouts as we have some automation for mirroring images to local registries that relies on the fully qualified path.

manifests/Makefile
```diff
@@ -0,0 +1,38 @@
+ARGO_ROLLOUTS_VERSION := v0.10.2
+
+TEMP_DIR := $(shell mktemp -d)
+
+.PHONY: argo-rollouts
+argo-rollouts: argo-rollouts-crds argo-rollouts-base
+
+.PHONY: argo-rollouts-crds
+argo-rollouts-crds:
```
moshloop, line 9: Maybe we should make this more generic e.g. `make upstreams` that looks at each file in the upstream directory and then renames it to `kustomization.yaml` and places the result in the file with the name ?

```bash
Makefile
upstream/
    crd/
        argo-rollouts.yaml
    argo-rollououts.yaml
```

manifests/Makefile
tobernguyen: Done

manifests/argo-rollouts.yaml
```diff
@@ -0,0 +1,291 @@
+apiVersion: v1
```
moshloop, line 1: Can we automatically insert warning headers that are generated by make ?

manifests/Makefile
```diff
@@ -0,0 +1,39 @@
+ARGO_ROLLOUTS_VERSION := v0.10.2
```
moshloop, line 1:
```suggestion
ARGO_ROLLOUTS_VERSION := $(shell yq r ../test/cicd.yaml argoRollouts.version)
```

manifests/argo-rollouts.yaml
tobernguyen: Yeah, I think we can achieve that at `make` level (use `sed` I guess). Looks like `kustomize` doesn't support adding a header to the build result.

manifests/Makefile
tobernguyen: Yeah, I like that. So:
1. Create a new directory `manifests/upstream`
2. Move `manifests/crd/argo-rollouts.yaml` to `manifests/upstream/crd`
3. Move `manifest/argo-rollouts.yaml` to `manifests/upstream`
4. Modify `Makefile` and make it generic by creating a new task that iterates over files in `upstream` then for each file: reads the file name then runs the task that has the same name with the file name (without the file extension)
5. Other tasks remain the same

Sounds right? I'm not sure about this part
> then renames it to kustomization.yaml and places the result in the file with the name 

We will apply this pattern for `argo-rollouts` first and going forward we will move other production runtimes manifests to follow this yaml, right?.

moshloop: 5. So it would go
`mv manifests/upstream/argo-rollouts.yaml $TMP/kustomization.yaml`
`kustomize build $TMP/ > manifests/argo-rollouts.yaml`

moshloop: And yes we would apply this pattern going forward, and then also backport to existing runtimes.

manifests/argo-rollouts.yaml
tobernguyen: This is done

manifests/Makefile
tobernguyen: After some changes to the Makefile, now I think it is better to use `manifests/upstream` as a single source of truth for generating manifests. So I'm hard-coding version of Argo Rollouts manifests in `upstream/argo-rollouts.yaml` and `upstream/crd/argo-rollouts.yaml`.

manifests/Makefile
tobernguyen: This is done. The `Makefile` is now much more elegant (I think 😄 ). It will rebuild manifests only if the corresponding upstream file is updated from the last time the manifest was built.

moshloop:
It looks like new tests added are not working / flakey ?

tobernguyen:
> It looks like new tests added are not working / flakey ?

Looks like it is fixed now (I increased the timeout time for the test).

PR #694, opened 2021-02-17
feat: add logs-exporter

teodor-pripoae:
Fixes: #691


pkg/phases/elasticsearch/install.go
```diff
@@ -26,5 +28,15 @@ func Deploy(p *platform.Platform) error {
 		p.Elasticsearch.Replicas = 3
 	}
 
+	if !p.LogsExporter.IsDisabled() {
+		if err := p.ApplySpecs(constants.PlatformSystem, "logs-exporter.yaml"); err != nil {
```
moshloop, line 32: It should still be possible to deploy logs exporter with elastic search configured elsewhere or even externally

PR #721, opened 2021-02-23
feat: add letsencrypt issuer

paddatrapper:


manifests/letsencrypt-issuer.yaml
```diff
@@ -0,0 +1,62 @@
+{{ if eq .dns.type "route53" }}
```
moshloop, line 1: We try and avoid templating entire documents, so in this case it would be better to move this to install.go

pkg/phases/letsencrypt/install.go
```diff
@@ -0,0 +1,11 @@
+package letsencrypt
+
+import "github.com/flanksource/karina/pkg/platform"
+
+func Install(p *platform.Platform) error {
+	if p.Letsencrypt.IsDisabled() {
```
moshloop, line 6: This would disable cert-manager which we don't want

moshloop: Nevermind I was misreading this

pkg/types/config.go
```diff
@@ -41,6 +41,7 @@ type PlatformConfig struct {
 	Kubernetes            Kubernetes          `yaml:"kubernetes" json:"kubernetes,omitempty"`
 	Kpack                 Kpack               `yaml:"kpack,omitempty" json:"kpack,omitempty"`
 	Ldap                  *Ldap               `yaml:"ldap,omitempty" json:"ldap,omitempty"`
+	Letsencrypt           LetsencryptIssuer   `yaml:"letsencrypt,omitempty" json:"letsencrypt,omitempty"`
```
moshloop, line 44: this should probably come under certmanager, as it doesn't make sense on its own

pkg/types/types.go
```diff
@@ -1074,3 +1075,8 @@ type ConfigDirective struct {
 	SopsPath   string             `yaml:"sops,omitempty" json:"sops,omitempty"`
 	SecretPath v1.SecretReference `yaml:"secretRef,omitempty" json:"secretRef,omitempty"`
 }
+
+type LetsencryptIssuer struct {
+	Disabled `yaml:",inline" json:",inline"`
+	Email    string `yaml:"email,omitempty" json:"email,omitempty"`
```
moshloop, line 1081: I think we can add the URL in here as well, defaulting to the prod API

manifests/letsencrypt-issuer.yaml
```diff
@@ -0,0 +1,62 @@
...
+          region: "{{.dns.region}}"
+          hostedZoneID: "{{.dns.zone}}"
+          accessKeyID: "{{.dns.accessKey}}"
+          secretAccessKeySecretRef:
+            name: prod-route53-credentials
+            key: secret-access-key
+      {{ else }}
+      http01:
+        ingress:
+          class: nginx
```
moshloop, line 35: The nginx deployment isn't using ingress classes - it is just the default ingress

manifests/letsencrypt-issuer.yaml
```diff
@@ -0,0 +1,62 @@
...
+  name: prod-route53-credentials
+type: Opaque
+stringData:
+  secret-access-key: "{{.dns.secretKey}}"
+{{ end }}
+---
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt-prod
```
moshloop, line 15: This would need to instead of the `ingress-ca` issuer as that is the one used by default when signing ingress certs, so this probably needs to move into the cert-manager/install.go and create the issuer programmatically.

PR #738, opened 2021-03-02
Karina config import secrets

### LINKED ISSUE #641: importSecrets
e.g. 

```yaml
importSecrets:
   - name: secret
     namespace: namesoace
```

For each value:

1) First try treat it as an `-e` path
2) Then try unmarshal based on extension (json, yml, yaml) and use as normal config

teodor-pripoae:
Fixes #641


moshloop:
@teodor-pripoae  can you check the test errors ?

PR #756, opened 2021-03-10
chore: set filebeat ssl default to disabled

BrendanGalloway:
### Description

Set the default filebeat logstash config to ssl.enbabled=false

The correct (todo) fix is to expand logstash config and filebeat templating to allow pass ssl settings through to the configmap.

### Dependencies

NA

### Breaking Change

- [ ] Yes
- [ X ] No

### Testing Notes

Verified in working config in customer cluster

### **Links**

NA


moshloop:
I think let's expand the config

Issue #462, opened 2020-08-07
karina db backup list


moshloop:
Print a list of backups saves to S3.

e.g. 

`karina db backup list`
```
URL                                                 AGE
s3://bucket/path/backup-tar.gz  10d
```

And `-q` and `-n` option to use like: 
`karina db restore $(karina db backup list -q -n 1)`

PR #762, opened 2021-03-12
feat: Added db backup list command

prasanthmj:
### Description
Refer issue #462 
Adds db backup list command

It also implements the flags -q and -n
-q for just listing the backups and -n to limit the number of lines output

### Dependencies

none

### Breaking Change

- [ ] Yes
- [x ] No

### Testing Notes

**What I did:**
Tested in local kind setup with actual S3 backup

create backups by running
karina db backup --name mycluster -c karina.yaml

Then list the backups using the command
karina db backup list --name mycluster -c karina.yaml

### **Links**

https://github.com/flanksource/karina/issues/462


pkg/client/postgres/tasks.go
```diff
@@ -95,9 +98,40 @@ func (db *PostgresDB) ScheduleBackup(schedule string) error {
...
+
+	if !quiet {
+		fmt.Printf("\nURL\tAGE\n")
+	}
+	for idx, o := range objs {
+		url := fmt.Sprintf("s3://%s/%s", op.LogicalBackup.S3Bucket, o.Key)
+		if quiet {
+			fmt.Println(url)
+		} else {
+			age := time.Now().Sub(o.LastModified)
```
moshloop, line 114: Can you change this to time.Since ?

pkg/client/postgres/tasks.go
```diff
@@ -95,9 +98,40 @@ func (db *PostgresDB) ScheduleBackup(schedule string) error {
 	return db.client.Apply(db.Namespace, job)
 }
 
-func (db *PostgresDB) ListBackups() ([]string, error) {
-	var backups []string
-	return backups, nil
+func (db *PostgresDB) ListBackups(quiet bool, num int) error {
```
moshloop, line 101: This function should return a slice of strings and an error, the printing to the console should happen in the cmd package

pkg/client/postgres/tasks.go
```diff
@@ -95,9 +98,40 @@ func (db *PostgresDB) ScheduleBackup(schedule string) error {
...
+	op := db.op.Configuration
+
+	objs := listObjects(db.s3, op.LogicalBackup.S3Bucket, db.Name)
+
+	if !quiet {
+		fmt.Printf("\nURL\tAGE\n")
```
moshloop, line 107: Can we add in size?

This should also use a tabwriter rather than inline tabs which are not going to handle differences in value length

pkg/client/postgres/tasks.go
prasanthmj: Will update

pkg/client/postgres/tasks.go
prasanthmj: can't return just slice of strings but can be a slice of structs since we will need the age, size etc

pkg/client/postgres/tasks.go
```diff
@@ -95,9 +97,26 @@ func (db *PostgresDB) ScheduleBackup(schedule string) error {
...
-	return backups, nil
+type BackupItem struct {
+	URL          string
+	LastModified time.Time
+	Size         int64
+}
+
+func (db *PostgresDB) ListBackups() ([]*BackupItem, error) {
+	op := db.op.Configuration
+	doneCh := make(chan struct{})
```
moshloop, line 108: we dont really use channels for context anywhere else - Could we switch it to https://pkg.go.dev/github.com/minio/minio-go/v7#Client.ListObjects  and then pass the context through as a param ?

We probably also want to sort by LastModified desc

cmd/db.go
```diff
@@ -1,9 +1,16 @@
 package cmd
 
 import (
+	"fmt"
+	"os"
+	"text/tabwriter"
+	"time"
+
+	"code.cloudfoundry.org/bytefmt"
```
moshloop, line 9: We try and follow the principle of a little copy and paste is better than a little dependency = Could we copy the files out of these repos (ensure the correct license header & copyright owner is listed) and then make them available as :

flanksource/commons/text.HumanSize and text.HumanDuration ?

pkg/client/postgres/tasks.go
prasanthmj: At present it is using minio-go/v6 where ListObjects doesn't take context as param. ( https://pkg.go.dev/github.com/minio/minio-go/v6#Client.ListObjects ) . Should we upgrade to  minio-go/v7 ? 
I will add sort by LastModified desc

tobernguyen:
I will take over this issue as the backup mechanism has been changed significantly and the method of listing backup using MinIO CLI won't work anymore.

tobernguyen:
The new implementation for listing backup is pushed here: https://github.com/flanksource/karina/pull/786/commits/13d48b5dde950cc15a87b4a1253a8d774e21d9ea Thanks @prasanthmj for your code. I took some of your code from this PR.

PR #767, opened 2021-03-17
Add harbor details to vsphere tests

paddatrapper:


test/vsphere/vsphere.yaml
```diff
@@ -63,3 +63,7 @@ s3:
   skipTLSVerify: true
   e2e:
     minio: false
+platformOperator:
```
moshloop, line 66: I think we want to move this into a self-hosted.yaml, and then pass through -c self-hosted.yaml as an environment variable into the self-hosted tests.  The vsphere yaml is meant for E2E tests on vsphere, not the E2E tests on kind, running inside vsphere

PR #768, opened 2021-03-17
feat: add logvolume alert to default prometheus rules

BrendanGalloway:
### Description

Adds a record of the weekly average log voume, and an alert if the current volume exceeds this by a factor of 2

### Dependencies

NA

### Breaking Change

- [ ] Yes
- [ X ] No

### Testing Notes

Expression logic verified in PromQL

### **Links**

NA

### DIFF
--- a/manifests/monitoring/alertmanager-rules.yaml.raw
+++ b/manifests/monitoring/alertmanager-rules.yaml.raw
@@ -204,6 +204,24 @@ spec:
             cluster:node_cpu_seconds_total:rate5m / count(sum(node_cpu_seconds_total)
             BY (instance, cpu))
           record: cluster:node_cpu:ratio
+    - name: log-count-record
+      interval: 1d
+      rules:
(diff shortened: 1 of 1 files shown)


manifests/monitoring/alertmanager-rules.yaml.raw
```diff
@@ -204,6 +204,20 @@ spec:
             cluster:node_cpu_seconds_total:rate5m / count(sum(node_cpu_seconds_total)
             BY (instance, cpu))
           record: cluster:node_cpu:ratio
+    - name: log-counts
+      rules:
```
moshloop, line 208: I think the recording rule and the alerting rule will need to be split, the recording should only happen every 24h and save into :elastic_documents_by_namespace_cluster_node:1d while the alerting should happen at a regular interval. 

The alert can then do an average over just the 7 samples

Can we also add the current and historic rate into the label?

PR #776, opened 2021-03-23
docs: Tidying Troubleshooting doc

flanksaul:
### Description

A tidy of the `admin-guide/troubleshooting.md `documentation.
CI fixes make it possible to bypass "required" checks manually. Unfortunately, GitHub Action has poor support for conditional job level logic and it's not possible to have a job be "required" and also to skip it when changes unrelated to core products are added. At present, the required checks will only be excluded a) if a skip-e2e string is in the commit message (this functionality has been fixed) and b) if the only changes are to the `docs` folder.

### Dependencies

None

### Breaking Change

- [ ] Yes
- [ x] No

### Testing Notes

Renders correctly with local build.

**What I did:**
Ran a local instance of mkdocs to verify changes visually.
Have tested various parameters to ensure that the logic works for skip-e2e and for docs exclusion 

Run a local instance of mkdocs to verify changes visually.
Use "skip-e2e" as a string in a commit message or change the value of "fileChangesToCoreArtifacts" to false and repush.

### **Links**


docs/overrides/icons/kubernetes.svg
```diff
@@ -1 +1,6 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" role="img" viewBox="-0.17 0.08 230.10 223.35">
```
flanksaul, line 1: Not sure why but the text in the `<style>` tags was rendering. Have just moved the fill to the paths.

docs/admin-guide/troubleshooting.md
```diff
@@ -93,114 +88,89 @@ The steps below are applicable for Sev 1 & 2 incidents, see the Escalations sect
...
+| Deployments                   | See [Troubleshooting Deployments](https://learnk8s.io/troubleshooting-deployments) <br>:bash: [kubectl-debug](https://github.com/aylei/kubectl-debug)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
flanksaul, line 126: My general take is that we should avoid `<a>` tags unless the result cannot be achieved in raw markdown (for example certain table formats). I have converted the links to standard markdown.

docs/admin-guide/troubleshooting.md
```diff
@@ -93,114 +88,89 @@ The steps below are applicable for Sev 1 & 2 incidents, see the Escalations sect
...
-
-
 ## Escalations
 
-After receiving a new alert or ticket, spend a few minutes (&lt; 5m) doing a preliminary investigation to identify the reproducibility and impact of the incident. Once confirmed as an issue being the formal incident response process:
-
-
-
-## Hypothesis development
+After receiving a new alert or ticket, spend a few minutes (&lt; 5m) do a preliminary investigation to identify the reproducibility and impact of the incident. Once confirmed as an issue, the formal incident response process is:
```
flanksaul, line 93: My reading here is that the headings between "Escalations" and "Incident Resolution" are part of the "Escalations" subsection, but not sure.

docs/admin-guide/troubleshooting.md
```diff
@@ -1,35 +1,30 @@
...
+| :captain: Incident Commander   | Communicating with stakeholders |
+| :talk: Incident Communicator   | Overseeing the incident         |
 
+The steps below are applicable for Severity ("**Sev**") 1 & 2 level incidents. See the [Escalations](#escalations) section for more detailed steps with times. See [Severity Classification](#severity-classification) for a severity classification rubric.
 
-
-The steps below are applicable for Sev 1 & 2 incidents, see the Escalations section for more detailed steps with times etc..
-
-<span style="font-size: 28px; color: grey">:1:</span> The first responder automatically becomes the incident commander until it is transferred:      <br/>
+:1: The first responder automatically becomes the incident commander until the role is transferred:
```
flanksaul, line 12: Re numbering: My general view is that we should avoid HTML in markdown where markdown is sufficient. Personally like the style without the tags, but even so think it's best to keep the MD as clean as possible and style with a stylesheet if necessary. 

Note: Numbering like this is something that would be checked in an accessibility audit; there are ways to make SVG numbering accessible though.

.github/workflows/kind.yml
```diff
@@ -1,51 +1,95 @@
...
     runs-on: ubuntu-20.04
+    needs: gates
     container:
       image: flanksource/build-tools:v0.12.0
+    env: 
+      FILE_CHANGES_TO_CORE_ARTIFACTS: ${{ needs.gates.outputs.fileChangesToCoreArtifacts == 'true' }}
     steps:
       # Get values for cache paths to be used in later steps
       - id: go-cache-paths
+        if: ${{ env.FILE_CHANGES_TO_CORE_ARTIFACTS == 'true' }}
```
paddatrapper, line 54: Is it possible to have a step that essentially is `if ${{ env.FILE_CHANGES_TO_CORE_ARTIFACTS == 'false' }}` then succeed without running the other steps?

flanksaul: Not that's I've been able to find. My first impulse was to try early return but they don't seem to operate that way (I could be wrong) and it seems without further research I think the idea is that that they're independent without explicitly stating deps.

paddatrapper: Ok. That is slightly annoying on GA's part...

PR #784, opened 2021-03-24
chore: add resource requests/limits to harbor

paddatrapper:
Fixes: #770

### DIFF
--- a/manifests/harbor/exporter.yaml
+++ b/manifests/harbor/exporter.yaml
@@ -77,7 +77,7 @@ spec:
             readOnlyRootFilesystem: true
           resources:
             limits:
-              cpu: 500m
+              cpu: 200m
               memory: 256Mi
             requests:
               cpu: 100m
--- a/manifests/harbor/exporter.yaml
+++ b/manifests/harbor/exporter.yaml
@@ -77,11 +77,11 @@ spec:
             readOnlyRootFilesystem: true
           resources:
             limits:
-              cpu: 400m
+              cpu: 500m
               memory: 256Mi
             requests:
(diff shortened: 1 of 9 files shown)


moshloop (changes requested):
Can we set minimum requests to 100m/200m cpu and 64/128Mi mem ?
Then when increasing them go in intervals of  

CPU: 100,200,500,1000,1500,2000
Mem: 64,128,256,512,1Gb,2gb,4gb etc..

Clair I think will need to burst to 1000m
Core and Registry cpu should be 200/500 and mem, 256/512
Registryctl can remain low 


e.g. 
clair

manifests/harbor/exporter.yaml
```diff
@@ -77,11 +77,11 @@ spec:
             readOnlyRootFilesystem: true
           resources:
             limits:
-              cpu: 400m
+              cpu: 500m
```
moshloop, line 80:
```suggestion
              cpu: 200m
```

moshloop (changes requested):
please check failing tests

moshloop (changes requested):
I think we might need to reduce the requests for the test run in https://github.com/flanksource/karina/blob/master/test/patch1.yaml as we are limited to 7GB of memory total

moshloop (changes requested):
So the requests/limits should have remained in manifests/harbor and then those new default requests should have been overridden in patch1, bring down the requirements for tests

moshloop:
@paddatrapper  can you check the failing harbor test ?

paddatrapper:
Harbor is now passing

PR #805, opened 2021-04-14
chore: add upstream canary checker

teodor-pripoae:


manifests/upstream/canary-checker/base/kustomization.yaml
```diff
@@ -0,0 +1,56 @@
+resources:
+  - https://raw.githubusercontent.com/flanksource/canary-checker/v0.19.0/config/deploy/base.yaml
+images:
+  - name: flanksource/canary-checker
+    newName: docker.io/flanksource/canary-checker
+    newTag: "{{.canaryChecker.version}}"
```
moshloop, line 6: I think this should have a default, especially for when trying to delete

teodor-pripoae: Ok, I've added the default value.

PR #807, opened 2021-04-14
chore: add cpu/memory fields to postgresqldb

teodor-pripoae:


moshloop:
@teodor-pripoae can this be merged ?

PR #816, opened 2021-04-19
feat: kpack support

Kaitou786:
Signed-off-by: Tarun Khandelwal <tarun@flanksource.com>

### Description

### Dependencies

- _Ex. Other PRs._

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

**What I did:**

Using the kpack.yaml config file for karina you can deploy kpack with `karing -c test/kpack.yaml deploy kpack`
### **Links**


pkg/phases/kpack/test.go
```diff
@@ -1,16 +1,62 @@
...
 	"github.com/flanksource/commons/console"
+	"github.com/flanksource/commons/exec"
 	"github.com/flanksource/karina/pkg/platform"
+	"github.com/flanksource/karina/pkg/types"
 	"github.com/flanksource/kommons"
 )
 
 func Test(p *platform.Platform, test *console.TestResults) {
-	if p.Kpack.IsDisabled() {
+	if p.Kpack.Disabled || (p.Kpack.ImageVersions == types.ImageVersions{}) {
```
moshloop, line 16: the blank image versions check should be moved inside IsDisabled()

pkg/phases/kpack/test.go
```diff
@@ -1,16 +1,62 @@
...
+
+func TestE2EKpack(p *platform.Platform, test *console.TestResults) {
+	testName := "kpack-e2e-test"
+	if err := p.ApplySpecs(Namespace, "kpack-e2e.yaml"); err != nil {
+		test.Failf(testName, "Failed to apply %v", err)
+		return
+	}
+	// Wait for Builder status to be true -> wait for 5 mintues
+	for i := 0; i < 5; i++ {
+		output, ok := exec.SafeExec(fmt.Sprintf("kubectl get builder test-builder -n %s", Namespace))
```
moshloop, line 35: When calling external binaries there is a helper in `platform.GetBinary`, and also a `platform.GetKubectl` that handles creation of a local kubeconfig - there might be a kubeconfig when using kind, but our E2E tests are designed to run in production (hence the integration into karina test and not via go test) so it can't assume a kubeconfig exists when it could be dynamically generated based on the master cert.

That being said we avoid calling external binaries because it tends to be slow and fragile, the  `platform.GetByKind` function is a direct replacement for `kubectl get`

In this case it shouldn't be necessary to use GetByKind either you can use `platform.WaitForResource` that does the looping and readiness checks already

Makefile
```diff
@@ -15,8 +15,8 @@ endif
 
 # Image URL to use all building/pushing image targets
 IMG ?= flanksource/karina:latest
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+# Produce CRDs with v1
+CRD_OPTIONS ?= "crd:crdVersions=v1"
```
moshloop, line 19: I know I said we could include the controller-gen changes in this PR, but it probably makes sense to split it out so it can merge faster. 

Even when including in the same PR, we would want to split `feat:` and `refactor:` commits

moshloop (changes requested):

moshloop:
It looks like the images are still using latest ?

2021-04-21 16:01:07 +0000 UTC    1      Normal   Pulling              Pod/test-image-build-1-5wpzl-build-pod    Pulling image   "gcr.io/cf-build-service-public/kpack/build-init:latest"

There is also a missing service account:


2021-04-21 16:00:44 +0000 UTC    1      Warning  FailedCreate       ReplicaSet/kpack-webhook-bf4df8cdc      Error creating: pods   "kpack-webhook-bf4df8cdc-" is forbidden: error looking up service   account kpack/webhook: serviceaccount "webhook" not found |  
-- | --

Kaitou786:
> It looks like the images are still using latest ?
> 

Oops, I forgot to update the test, they were using the latest in the defined configuration

> There is also a missing service account:

I am not sure... Seeing in the logs it seemed to be created: https://github.com/flanksource/karina/runs/2402084895#step:4:783

Kaitou786:
Looking at the logs with the new deployment test for webhook and controller the pod seems to be running: https://github.com/flanksource/karina/pull/816/checks?check_run_id=2408734773#step:4:1024

test/test.sh
```diff
@@ -93,6 +93,10 @@ fi
 
 $BIN version
 
+# Install genuinetools/reg
+sudo curl -fSL "https://github.com/genuinetools/reg/releases/download/v0.16.1/reg-linux-amd64" -o "/usr/local/bin/reg"
```
moshloop, line 97: Can we move this to: https://github.com/flanksource/commons/blob/master/deps/deps.go#L60

moshloop (changes requested):

moshloop (approved):

PR #821, opened 2021-04-21
feat: ingest Kubernetes audit logs

paddatrapper:
### Description

Fixes: #597

### Breaking Change

- [ ] Yes
- [x] No

### DIFF
--- a/manifests/filebeat.yaml
+++ b/manifests/filebeat.yaml
@@ -174,6 +174,15 @@ data:
       hosts: ["${LOGSTASH_URL}"]
       ssl: {{ .ssl | default ( coll.Dict "enabled" "false" ) | data.ToJSON }}
 {{ end }}
+{{ if eq .name "infra" }}
+{{ if ".kubernetes.apiServerExtraArgs.audit-log-path" }}
+    filebeat.inputs:
+      - type: log
+        enabled: true
+        paths:
+          - '{{ ".kubernetes.apiServerExtraArgs.audit-log-path" }}'
+{{ end }}
+{{ end }}
 ---
 {{end}}
 apiVersion: rbac.authorization.k8s.io/v1


manifests/filebeat.yaml
```diff
@@ -174,6 +174,13 @@ data:
       hosts: ["${LOGSTASH_URL}"]
       ssl: '{{ .ssl | default ( coll.Dict "enabled" "false" ) | data.ToJSON }}'
 {{ end }}
+{{ if ".kubernetes.apiServerExtraArgs.audit-log-path" }}
```
moshloop, line 177: Whats the difference between `if ".kubernetes.apiServerExtraArgs.audit-log-path"` and `if .kubernetes.apiServerExtraArgs.audit-log-path` ?

I think this might also need to be scoped only to the infra-filebeat, not the other custom ones

paddatrapper: The latter throws a malformed yaml error when trying to be parsed. See https://github.com/flanksource/karina/runs/2400317152?check_suite_focus=true#step:4:736

Ok, how do I go about doing that?

moshloop: > The latter throws a malformed yaml error when trying to be parsed.

Ah thats due to the "-"  in audit-log-path

`if and .name == "infra" ".kubernetes.apiServerExtraArgs.audit-log-path" `

manifests/filebeat.yaml
```diff
@@ -174,6 +174,15 @@ data:
       hosts: ["${LOGSTASH_URL}"]
       ssl: '{{ .ssl | default ( coll.Dict "enabled" "false" ) | data.ToJSON }}'
 {{ end }}
+{{ if eq .name "infra" }}
+{{ if ".kubernetes.apiServerExtraArgs.audit-log-path" }}
+    filebeat.inputs:
+      - type: log
+        enabled: true
+        paths:
+          - '{{ ".kubernetes.apiServerExtraArgs.audit-log-path" }}'
```
paddatrapper, line 183: There is an issue with this - the path variable replacement doesn't work and produces:

```yaml
filebeat.inputs:
  - type: log
    enabled: true
    paths:
      - '.kubernetes.apiServerExtraArgs.audit-log-path'
```

This is because the quoting required to make the variable `.kubernetes.apiServerExtraArgs.audit-log-path` valid yaml seems to be interpreted as a string by our templating system and not as a quoted variable.

However I am not sure how to fix it. All permutations of quoting result in either invalid yaml or some variation of the result above

PR #832, opened 2021-04-27
chore: Lint duplicate manifests

Kaitou786:
Signed-off-by: Tarun Khandelwal <tarun@flanksource.com>

### Description

Check all the manifests and see if there is any duplicate keys found the identification key is in the form `<kind>-<name>-<namespace>` or `<kind>-<name>` in case namespace is not present

### Dependencies

- _Ex. Other PRs._

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

**What I did:**
Just run the linter file directly `go run test/linter/main.go`

### **Links**

### DIFF
--- a/test/linter/main.go
+++ b/test/linter/main.go
@@ -8,6 +8,8 @@ import (
 	"path/filepath"
 	"regexp"
 
+	"github.com/flanksource/kommons"
+
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/flanksource/yaml.v3"
@@ -17,8 +19,19 @@ const (
 	configFile = "test/linter.yaml"
 )
 
+var allowedDuplicateKeys = []string{"CustomResourceDefinition-servicemonitors.monitoring.coreos.com"}
+var ignoreManifestsSubPaths = []string{
+	"manifests/gatekeeper/(.*)",
+	"manifests/harbor/(.*)",
+	"manifests/monitoring/(.*)",
+	"manifests/nginx/(.*)",
+	"manifests/template/(.*)",
+	"manifests/test/(.*)",
+	"manifests/upstream/(.*)",
+}
 var (
 	config = &Config{}
+	keys   = make([]string, 10, 500)
 )
 
 type Config struct {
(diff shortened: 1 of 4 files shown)


test/linter/main.go
```diff
@@ -17,8 +19,19 @@ const (
 	configFile = "test/linter.yaml"
 )
 
+var allowedDuplicateKeys = []string{"CustomResourceDefinition-servicemonitors.monitoring.coreos.com"}
+var ignoreManifestsSubPaths = []string{
```
moshloop, line 23: only the upstream should be skipped

BrendanGalloway:
@Kaitou786 would you mind putting in a line or two for description in the cover sheet, so that reviewers have a bit of context as to what the change is intended to do.

test/linter/main.go
```diff
@@ -49,6 +54,57 @@ func checkPath(filePath string, info os.FileInfo, err error) error {
...
+	}
+	manifestData, err := kommons.GetUnstructuredObjects(yamlFile)
+	if err != nil {
+		log.Warnf("error parsing the yaml %v", manifest)
+	}
+	for i := range manifestData {
+		if manifestData[i].Object["metadata"].(map[string]interface{})["namespace"] != nil {
+			value := fmt.Sprintf("%v-%v-%v", manifestData[i].Object["kind"].(string), manifestData[i].Object["metadata"].(map[string]interface{})["name"].(string), manifestData[i].Object["metadata"].(map[string]interface{})["namespace"].(string))
+			if !contains(allowedDuplicateKeys, value) && contains(keys, value) {
+				return errors.Errorf("error %v already present in the manifests", value)
```
BrendanGalloway, line 94: Could the error also state which file the duplicate is encountered in?

Kaitou786:
> @Kaitou786 would you mind putting in a line or two for description in the cover sheet, so that reviewers have a bit of context as to what the change is intended to do.

Ah sure I will from now on.

test/linter/main.go
Kaitou786: Thinking of a way it could be possible.

Kaitou786: Changed keys to a map with map-key as the generated key and value as the manifest file. You can test by edit the allowedDuplicateManifests var to any other value and see the test fails with giving msg like:
```
FATA[0000] Manifest linting failed: error: key CustomResourceDefinition-servicemonitors.monitoring.coreos.com from manifests/crd/service-monitor.yaml already present in manifests/crd/prometheus.yaml
```

BrendanGalloway (approved):

test/linter/main.go
BrendanGalloway: Thanks, that is a nice level of verbosity

Issue #840, opened 2021-04-29
Standardise enable/disable flags across platform config objects


BrendanGalloway:
Currently some platform configs expect to be nil if undefined, others have inline enable/disable flags and others have explicit enable/disable flags.  Flag naming also varies between enabled and disabled.  There should be a standard approach for all config objects

PR #841, opened 2021-04-29
feat: allow vsphere config to be explicitly disabled

BrendanGalloway:
### Description

Add a Disabled to the Vsphere config struct such that it can be explicitly disabled
This is necesary for cases wheree a config file with vsphere needs  to be merged into one without, eg to recreate a cluster with a different cni/csi combo.  Currently the vsphere phase checks for the vsphere config to be nil, but there is no way to insert a nil value into the config merge.

Removed the "Data" entry from platform config.  This was added to allow yaml anchors to be placed in config files, but there is no data type that supports both anchors and use in crds, so it will be removed instead.

### Dependencies

- _Ex. Other PRs._

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

**What I did:**

### **Links**

#840


pkg/types/config.go
```diff
@@ -23,7 +19,6 @@ type PlatformConfig struct {
 	Consul    string    `yaml:"consul" json:"consul,omitempty"`
 	Dashboard Dashboard `yaml:"dashboard,omitempty" json:"dashboard,omitempty"`
 	// Semistructured data to be reused using YAML anchors
```
moshloop, line 21: I think the comment also needs to be removed

PR #842, opened 2021-04-29
fix: ensure 'cloud' is called in phases loop

flanksaul:
### Description

Mismatch between `cloud` in the PhaseOrder array and `cloud-controller` at:https://github.com/flanksource/karina/blob/7417b14166e82d646622b25b029d4207611c1470/pkg/phases/order/order.go#L144

### Dependencies

- _Ex. Other PRs._

### Breaking Change

- [x] Yes
- [ ] No

### Testing Notes

Any consumers of `cloud-controller` may have issues, but I don't have enough context to know if this is important. @moshloop?

**What I did:**

### **Links**

### DIFF
--- a/pkg/phases/order/order.go
+++ b/pkg/phases/order/order.go
@@ -141,7 +141,7 @@ var PhasesExtra = map[string]DeployFn{
 	"quack":              quack.Install,
 	"template-operator":  templateoperator.Install,
 	"vsphere":            vsphere.Install,
-	"cloud-controller":   Cloud,
+	"cloud":              Cloud,
 	"stubs":              Stubs,
 }
 


pkg/phases/order/order.go
```diff
@@ -141,7 +141,7 @@ var PhasesExtra = map[string]DeployFn{
 	"quack":              quack.Install,
 	"template-operator":  templateoperator.Install,
 	"vsphere":            vsphere.Install,
-	"cloud-controller":   Cloud,
+	"cloud":              Cloud,
```
BrendanGalloway, line 144: To avoid the possibility of this being a breaking change, this line could be an addition instead of a replacement.  That said I don't believe there's anywhere we currently call `cloud-controller` in a script and anyone currently calling `cloud` would result in a dictionary key lookup failing at https://github.com/flanksource/karina/blob/7417b14166e82d646622b25b029d4207611c1470/cmd/deploy.go#L34 so I would be fine with this merging as it.

PR #862, opened 2021-05-06
Setup cert-manager certificates for harbor-trivy

### LINKED ISSUE #854: Harbor - Trivy scan errors due to missing docker socket

BrendanGalloway:
### Description

Add certificates, volumemounts and volume for harbor-trivy

This should be the actual fix for #854 - tracing through the trivy logic leads to https://github.com/aquasecurity/fanal/blob/main/image/image.go#L70 which shows trivy attempts to read the image from a local docker registry via socket, from podman and then from a remote registry.  The third method is the desired usecase in harbor and the error regarding connection to a docker socket is misleading.  The error regarding unknown CA is the actual error, which is addressed here.

### Dependencies

NA

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

Applied to kind cluster with harbor2.yaml deployed.

before patch, connection from trivy to harbor core fails:
```
sh-4.4$ curl https://harbor-core.harbor.svc.cluster.local.:443/v2/
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```
After patch:
```
sh-4.4$ curl https://harbor-core.harbor.svc.cluster.local.:443/v2/
{"errors":[{"code":"UNAUTHORIZED","message":"unauthorized: unauthorized"}]}
```

### **Links**

Fixes #854

### DIFF
--- a/manifests/harbor/core.yaml
+++ b/manifests/harbor/core.yaml
@@ -33,7 +33,7 @@ data:
   CFG_EXPIRATION: "5"
   ADMIRAL_URL: "NA"
   WITH_TRIVY: "true"
-  TRIVY_ADAPTER_URL: "http://harbor-trivy.harbor.svc.cluster.local.:8080"
+  TRIVY_ADAPTER_URL: "https://harbor-trivy.harbor.svc.cluster.local.:8080"
   REGISTRY_STORAGE_PROVIDER_NAME: "s3"
   WITH_CHARTMUSEUM: "true"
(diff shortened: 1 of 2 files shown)


moshloop:
@BrendanGalloway can you check the failing harbor test ?

PR #877, opened 2021-05-16
fix: change DeleteSpecs to check if core api type resources exist

teodor-pripoae:


pkg/constants/const.go
```diff
@@ -40,3 +40,26 @@ var PlatformNamespaces = []string{
 	"vault",
 	"velero",
 }
+
```
moshloop, line 43: Can we move this into Kommons ?

moshloop (changes requested):
can we move the IsCoreType to kommons ?

PR #879, opened 2021-05-17
feature: add field to configure nginx ingress to not use host network

teodor-pripoae:


manifests/nginx.yaml
```diff
@@ -174,7 +174,9 @@ spec:
         - name: scripts
           configMap:
             name: lua-scripts
+      {{ if not (.nginx.disableHostNetwork) }}
       hostNetwork: true
```
moshloop, line 178:
```suggestion
      hostNetwork: {{  ternary "false" "true" .nginx.disableHostNetwork }}
```

PR #883, opened 2021-05-20
fix: remove ingress class

teodor-pripoae:


paddatrapper:
This was added due to errors without a default ingress class specified on BCB infrastructure

moshloop:
@paddatrapper Yeh, but it didn;t resolve those errors and introduced new issues

PR #889, opened 2021-05-24
feat: process config paths as relative to config file, not karina

BrendanGalloway:
### Description

This change will process paths in config files relative to config file location rather than go wd, no longer requiring apriori knowledge of the location karina will be called from.  This should simplify long chains of config files, as well as make aligning scripting and config management easier.

### Dependencies

NA

### Breaking Change

- [x] Yes
- [ ] No

This change will break existing config files that assume karina will be called from some location other than the config file's location.  

### Testing Notes

Used karina to render some the yaml files in the test folder, checking that the output is the same whether karina is called from the root of the repo or within the test folder.  All paths in the rendered output point to the absolute location of the files in question

### **Links**

NA


cmd/common.go
```diff
@@ -173,6 +187,38 @@ func mergeConfigBytes(base *types.PlatformConfig, data []byte, path string) erro
 		return errors.Wrap(err, "Failed to parse YAML")
 	}
 
+	for k := range cfg.Patches {
+		if !strings.Contains(cfg.Patches[k], "\n") && (strings.HasSuffix(cfg.Patches[k], "yaml") || strings.HasSuffix(cfg.Patches[k], "yml")) {
```
moshloop, line 191: I think move this to a `IsValidPath(cfg.Patches[k], "yaml", "yml", "json")`

PR #897, opened 2021-06-03
feat: allow Connection to read secrets from running cluster directly

BrendanGalloway:
### Description

Change Connection.Password to an Kommons.EnvVar to allow karina to read elastic secrets directly.  Required for single-pass install of both elastic and filebeat

### Dependencies

NA

### Breaking Change

- [x] Yes
- [] No

Changes config file format, existing config files will need to be updated

### Testing Notes
Deployed a kind cluster with eck and set up an elastic instance.  Ran 'karina deploy filebeat -c test/elastic.yaml` with the currently commented 'valueFrom' field in the the config file uncommented (value commented out instead)

elastic-infra secret created with the correct credentials

### **Links**

NA


pkg/types/types.go
```diff
@@ -993,12 +995,12 @@ func (v Vsphere) GetSecret() map[string][]byte {
...
-	Password string `yaml:"password,omitempty" json:"password,omitempty"`
-	Port     string `yaml:"port,omitempty" json:"port,omitempty"`
-	Scheme   string `yaml:"scheme,omitempty" json:"scheme,omitempty"`
-	Verify   string `yaml:"verify,omitempty" json:"verify,omitempty"`
+	URL      string         `yaml:"url" json:"url"`
+	User     string         `yaml:"user,omitempty" json:"user,omitempty"`
+	Password kommons.EnvVar `yaml:"password,omitempty" json:"password,omitempty"`
+	Port     string         `yaml:"port,omitempty" json:"port,omitempty"`
+	Scheme   string         `yaml:"scheme,omitempty" json:"scheme,omitempty"`
+	Verify   string         `yaml:"verify,omitempty" json:"verify,omitempty"`
```
moshloop, line 1003: Probably makes sense to convert all of these to EnvVar

BrendanGalloway: I'm not sure - are the url/port/scheme likely to be in a configmap/secret?  In the case of elasticsearch the user is a key and not a value, so it can't be read from config either.  That might not be the case if we reuse connection elsewhere, but currently it's only used for *beat objects.

Issue #896, opened 2021-06-03
karina version includes only timestamp


moshloop:
e.g.

-20210602211129

it should be
v0.45.0-20210602211129

PR #922, opened 2021-06-25
fix: karina version includes only timestamp #896

apaarshrm39:
### Description

### Dependencies

- _Ex. Other PRs._

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

**What I did:** Fixed the fetch depth in GitHub workflow's checkout action, so that tags can be retrieved 
Checked  the Workflow output's in the fork 
image

### **Links**

#896 review-requested: @moshloop

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,11 @@
 default: build
 NAME:=karina
 
-VERSION_TAG=$(VERSION)-$(shell date +"%Y%m%d%H%M%S")
+ifeq ($(VERSION),)
+  VERSION_TAG=$(shell git describe --abbrev=0 --tags --exact-match 2>/dev/null || git describe --abbrev=0 --all)-$(shell date +"%Y%m%d%H%M%S")
+else
+  VERSION_TAG=$(VERSION)-$(shell date +"%Y%m%d%H%M%S")
+endif
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -25,7 +29,7 @@ release: linux darwin compress
 
 .PHONY: build
 build:
-	go build -o ./.bin/$(NAME) -ldflags "-X \"main.version=$(VERSION)\""  main.go
+	go build -o ./.bin/$(NAME) -ldflags "-X \"main.version=$(VERSION_TAG)\""  main.go
 
 
 .PHONY: linux
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -55,7 +55,9 @@ jobs:
         run: |
           echo "::set-output name=go-build::$(go env GOCACHE)"
           echo "::set-output name=go-mod::$(go env GOMODCACHE)"
-      - uses: actions/checkout@main
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
         if: ${{ env.FILE_CHANGES_TO_CORE_ARTIFACTS == 'true' }}
       # Cache go build cache, used to speedup go test
       - name: Go Build Cache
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -2,7 +2,11 @@
 default: build
 NAME:=karina
 
-VERSION_TAG=$(VERSION)-$(shell date +"%Y%m%d%H%M%S")
+ifeq ($(VERSION),)
+  VERSION_TAG=$(shell git describe --abbrev=0 --tags)-$(shell date +"%Y%m%d%H%M%S")
```
BrendanGalloway, line 6: For sanity when handling multiple local builds could I suggest:
```suggestion
  VERSION_TAG=$(shell git describe --abbrev=0 --tags --exact-match 2&>/dev/null || git describe --abbrev=0 --all)-$(shell date +"%Y%m%d%H%M%S")
```
this will set VERSION to the name of the branch being built in when the commit isn't exactly release-tagged.

@moshloop I see there's also main.date and main.commit vars, would it make more sense to set main.date ldflag rather than putting a date string in the version information?

apaarshrm39: @BrendanGalloway Thanks for the suggestion, I committed the same for now :)

apaarshrm39: @BrendanGalloway I tried your suggestion and unfortunately that again skips the release tags, reference the Image below. 
image

Maybe we can use "--always" flag as the fallback when tags are not available. 
image

BrendanGalloway: @apaarshrm39  - the idea is less that tags would not be available, but more that when working in local repos the most recent tag in history will be picked up, which would result in dev builds having (incorrect) release version info

It looks like you accidentally inserted a space between `&` and `>` in the version you tested (you've amended it out so I can't be sure):

```bash
~/Flanksource/karina(testenv)$ sed -n 6p Makefile 
        VERSION_TAG=$(shell git describe --abbrev=0 --tags --exact-match 2& >/dev/null || git describe --abbrev=0 --all)-$(shell date +"%Y%m%d%H%M%S")
~/Flanksource/karina(testenv)$ make
fatal: Not a valid object name 2
go build -o ./.bin/karina -ldflags "-X \"main.version=-20210625170658\""  main.go
~/Flanksource/karina(testenv)$ #Fix space in makefile
~/Flanksource/karina(testenv)$ sed -n 6p Makefile 
        VERSION_TAG=$(shell git describe --abbrev=0 --tags --exact-match 2&>/dev/null || git describe --abbrev=0 --all)-$(shell date +"%Y%m%d%H%M%S")
~/Flanksource/karina(testenv)$ make
go build -o ./.bin/karina -ldflags "-X \"main.version=heads/testenv-20210625170711\""  main.go
```
the `invalid object name 2` indicates 2 was passed to git as an argument, rather than used as a redirect for stderr

apaarshrm39: @BrendanGalloway Thanks, I don't believe  there was any spaces as I just committed your suggestion. Plus I rechecked it again.

 It was something with 2&>/dev/null itself.

To test this I tried to redirect the output to a file instead of /dev/null and see : 
image

The "2&>" command is busted anyway, it's that the shell used by the makefile doesn't suppress errors
One alternative to make the above command work anyway is to change the shell in makefile to /bin/bash
but,
also I couldn't understand the reason for using 2&>/dev/null as 
1) 2 is used to redirect STDERR
2)& is used to redirect both STDERR and STDOUT  

it seems to be sending the STDOUDT to null to 
If we are just trying to send the error to /dev/null so I believe just 2 alone should do the trick.

BrendanGalloway: @apaarshrm39

You are correct, it should just be `2>` not `2&>`.  I see you sent that up in 88b1505 - does that work better on your side, or does it still print the error about 2 not being a valid object?

apaarshrm39: @BrendanGalloway, Yeah it work as intended on my side now ! Thank you !

BrendanGalloway (approved):

Issue #449, opened 2020-07-27
karina delete-failed


moshloop:
essentially create a command that replicases this bash functionality
```bash
kdeletefailed () {
	IFS=$(echo -en "\n\b")
	for po in $(kubectl get po --all-namespaces | grep -v Running )
	do
		kubectl delete po --wait=false -n $(echo $po | awk '{print $1}') $(echo $po | awk '{print $2}')
	done
}
```

moshloop:
+ Include failed jobs created by cronjobs.

PR #923, opened 2021-06-28
feat: add cleanup pod function #449

apaarshrm39:
### Description

deletes all the pods from all the namespaces which are not in running state:

image

### Dependencies

- _Ex. Other PRs._

### Breaking Change

- [ ] Yes
- [ ] No

### Testing Notes

**What I did:**

created a kind cluster and tested functionality.

### **Links**

#449 

review-requested: @moshloop


BrendanGalloway:
@apaarshrm39 we don't mind force pushes and amends in PRs (although ideally they should only happen when the change is ready to merge and it's just fixups being squashed).  In this case though, would you mind rebasing on only the commits that are relevant to the change onto master and pushing that up, rather than the irrelevant commits and another commit to revert them...

cmd/deletefailed.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/spf13/cobra"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// deletefailedCmd represents the deletefailed command
+var Deletefailed = &cobra.Command{
+	Use:   "deletefailed",
```
BrendanGalloway, line 30: To align with the 'karina cleanup jobs` could this be 'karina cleanup pods'?

Would need a small rework of the existing `cleanup` command to set `jobs` and `pods` each as subcommands.

cmd/deletefailed.go
```diff
@@ -0,0 +1,71 @@
...
+	Short: "Delete the non running Pods",
```
BrendanGalloway, line 31:
```suggestion
	Short: "Delete non running Pods",
```

cmd/deletefailed.go
```diff
@@ -0,0 +1,71 @@
...
+		p := getPlatform(cmd)
+		client, err := p.GetClientset()
+		if err != nil {
+			p.Fatalf("unable to get clientset: %v", err)
+		}
+
+		// gather the list of Pods from all
+		pods, err := client.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			p.Fatalf("Failed to gather the list of jobs from namespace %v: %v", namespace, err)
```
BrendanGalloway, line 42: you reference `namespace` in your error message, but the var has not been initialised.  In addition, you are not specifying a namespace in your search criteria when retrieving the pod list, so the error message would not make sense in context even if the var did exist.

cmd/deletefailed.go
```diff
@@ -0,0 +1,71 @@
...
+	Short: "Delete the non running Pods",
+	Run: func(cmd *cobra.Command, args []string) {
+		p := getPlatform(cmd)
+		client, err := p.GetClientset()
+		if err != nil {
+			p.Fatalf("unable to get clientset: %v", err)
+		}
+
+		// gather the list of Pods from all
+		pods, err := client.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
```
BrendanGalloway, line 40: If the intention is to search across all namespaces it's clearer to use `metav1.NamespaceAll` rather than `""`.  From a functionality point of view, I'd prefer the option to specify a namespace to run this commands against, possibly with a check that sets `namespace` to `NamespaceAll` if no argument is given.

cmd/deletefailed.go
```diff
@@ -0,0 +1,71 @@
...
+
+		for _, po := range pods.Items {
+			wg.Add(1)
+
+			go func(po v1.Pod, clientSet *kubernetes.Clientset) {
+
+				condition := po.Status.Phase
+
+				// if the Condition of the Pod is not running, delete the Pod
+				if condition != "Running" {
```
BrendanGalloway, line 56: I believe this step can be done by setting  `ListOptions.FieldSelector` to `"status.Phase!=Running"` in the initial pod list retrieval.  FieldSelector has a fairly limited set of filterable fields, pod it looks like pods support status.phase: https://github.com/flant/shell-operator/blob/8fa3c3b8cfeb1ddb37b070b7a871561fdffe788b/HOOKS.md#fieldselector

BrendanGalloway (changes requested):

cmd/deletefailed.go
apaarshrm39: @BrendanGalloway I tried to incorporate all the PR suggestions, Can you please review now ? and kindly suggest if any other changes are required. 
image

BrendanGalloway (approved):

BrendanGalloway:
@apaarshrm39 Excellent, this looks pretty good now.  Given the third commit rewrites the contents of the first two, please squash this PR down to a single commit and resubmit.  Also please check gofmt and golanci-lint the code.

apaarshrm39:
> @apaarshrm39 Excellent, this looks pretty good now. Given the third commit rewrites the contents of the first two, please squash this PR down to a single commit and resubmit. Also please gofmt and golanci-lint the code.

Squashed all the commits into one.

BrendanGalloway:
> > @apaarshrm39 Excellent, this looks pretty good now. Given the third commit rewrites the contents of the first two, please squash this PR down to a single commit and resubmit. Also please gofmt and golanci-lint the code.
> 
> Squashed all the commits into one and made the lint fixes.

Cool - would you mind just  fixing the name of the commit/PR - deleteFailed isn't a thing any more.  Something like `feat: add pod cleanup function`

BrendanGalloway (approved):

Issue #645, opened 2021-02-02
karina seal


moshloop:
Mirror the kubeseal seal command

PR #927, opened 2021-07-06
feat: add seal command for kubeseal

paddatrapper:
### Description

Adds support for generating sealedsecrets using kubeseal

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

**What I did:**
Replaced `kubeseal` with `karina seal` in calls to create sealedsecrets

### **Links**

Fixes: #645

### DIFF
--- a/cmd/seal.go
+++ b/cmd/seal.go
@@ -1,0 +1,106 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func getFlagString(name string, cmd *cobra.Command) string {
+	value, err := cmd.Flags().GetString(name)
+	if value == "" || err != nil {
+		return ""
+	}
+	return fmt.Sprintf("--%s %s", name, value)
+}
+
+func getFlagFilePath(name string, cmd *cobra.Command) string {
+	value, err := cmd.Flags().GetString(name)
+	if value == "" || err != nil {
+		return ""
+	}
+	filePath, err := os.Getwd()
+	if err == nil {
+		filePath = fmt.Sprintf("%s/%s", filePath, value)
+	} else {
+		filePath = value
+	}
+	return fmt.Sprintf("--%s %s", name, filePath)
+}
+
+func getFlagFilePathSlice(name string, cmd *cobra.Command) string {
+	argList, err := cmd.Flags().GetStringSlice(name)
+	if err != nil || len(argList) == 0 {
+		return ""
+	}
+	result := ""
+	filePath, err := os.Getwd()
+	if err == nil {
+		filePath = fmt.Sprintf("%s/", filePath)
+	} else {
+		filePath = ""
+	}
+	for _, value := range argList {
+		value = fmt.Sprintf("--%s %s/%s", name, filePath, value)
+		result = fmt.Sprintf("%s %s", result, value)
+	}
+	return fmt.Sprintf("--%s %s", name, result)
+}
+
+func getFlagBool(name string, cmd *cobra.Command) string {
+	value, _ := cmd.Flags().GetBool(name)
+	if !value {
+		return ""
+	}
+	return fmt.Sprintf("--%s", name)
+}
+
+var Seal = &cobra.Command{
+	Use:   "seal",
+	Short: "Seal a secret using sealed-secrets",
+	Args:  cobra.MinimumNArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		p := getPlatform(cmd)
+		flags := []string{
+			getFlagString("format", cmd),
+			getFlagFilePath("sealed-secret-file", cmd),
+			getFlagBool("fetch-cert", cmd),
+			getFlagBool("allow-empty-data", cmd),
+			getFlagBool("validate", cmd),
+			getFlagFilePath("merge-into", cmd),
+			getFlagBool("raw", cmd),
+			getFlagBool("re-encrypt", cmd),
+			getFlagString("name", cmd),
+			getFlagFilePathSlice("from-file", cmd),
+		}
+
+		if !p.SealedSecrets.Disabled && p.SealedSecrets.Certificate != nil {
+			if p.SealedSecrets.Certificate.Cert != "" {
+				flags = append(flags, "--cert", p.SealedSecrets.Certificate.Cert)
+			}
+		}
+
+		kubeseal := p.GetBinary("kubeseal")
+		flagString := strings.Join(flags, " ")
+		argString := strings.Join(args, " ")
+		if err := kubeseal(flagString + argString); err != nil {
+			log.Fatalf("failed to run kubeseal: %v", err)
+		}
+	},
+}
+
+func init() {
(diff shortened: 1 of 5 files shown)


cmd/seal.go
```diff
@@ -0,0 +1,113 @@
...
+		if err := kubeseal(flagString + argString); err != nil {
+			log.Fatalf("failed to run kubeseal: %v", err)
+		}
+	},
+}
+
+func init() {
+	Seal.Flags().String("cert", "", "Certificate/public key file/URL to use for encryption")
+	Seal.Flags().String("controller-namespace", metav1.NamespaceSystem, "Namespace of sealed-secrets-controller")
+	Seal.Flags().String("controller-name", "sealed-secrets-controller", "Name of sealed-secrets controller.")
```
moshloop, line 99: So these fields should be based on platform.SealedSecrets, the interface for the user would be:

`karina seal secret.yaml -c karina.yaml > sealed.yaml`
and 
`karina unseal sealed.yaml -c karina.yaml > secret.yaml`

cmd/unseal.go
```diff
@@ -0,0 +1,130 @@
...
+			os.Remove(privateKeyFile)
+			log.Fatalf("failed to run kubeseal: %v", err)
+		}
+		os.Remove(privateKeyFile)
+	},
+}
+
+func init() {
+	Unseal.Flags().StringP("format", "o", "json", "Output format for sealed secret. Either json or yaml")
+	Unseal.Flags().Bool("allow-empty-data", false, "Allow empty data in the secret object")
```
moshloop, line 128: I think we can remove this option

cmd/unseal.go
```diff
@@ -0,0 +1,130 @@
...
+		if err := kubeseal(flagString + argString); err != nil {
+			os.Remove(privateKeyFile)
+			log.Fatalf("failed to run kubeseal: %v", err)
+		}
+		os.Remove(privateKeyFile)
+	},
+}
+
+func init() {
+	Unseal.Flags().StringP("format", "o", "json", "Output format for sealed secret. Either json or yaml")
```
moshloop, line 127:
```suggestion
	Unseal.Flags().StringP("format", "o", "yaml", "Output format for sealed secret. Either json or yaml")
```

cmd/seal.go
```diff
@@ -0,0 +1,103 @@
...
+		flagString := strings.Join(flags, " ")
+		argString := strings.Join(args, " ")
+		if err := kubeseal(flagString + argString); err != nil {
+			log.Fatalf("failed to run kubeseal: %v", err)
+		}
+	},
+}
+
+func init() {
+	Seal.Flags().StringP("format", "o", "json", "Output format for sealed secret. Either json or yaml")
```
moshloop, line 93: If you move the declaration of Seal into the init function, then you can define the variables up front, and then use StringVar, BoolBar etc without the need for getFlag functions

paddatrapper: Those getFlag functions also add the required `--name` flags that are passed to the kubeseal function call. They would still be needed for the filepath and slice values, but could be removed for the others

paddatrapper: Moving the declaration also doesn't make the getFlagBool calls any simpler, as they need to conditionally add a flag based on the input.

cmd/unseal.go
```diff
@@ -0,0 +1,129 @@
...
+			log.Fatalf("Unable to write sealed-secrets-key to disk: %v", err)
+		}
+		kubeseal := p.GetBinary("kubeseal")
+		flagString := strings.Join(flags, " ")
+		argString := strings.Join(args, " ")
+		if err := kubeseal(flagString + argString); err != nil {
+			os.Remove(privateKeyFile)
+			log.Fatalf("failed to run kubeseal: %v", err)
+		}
+		os.Remove(privateKeyFile)
```
moshloop, line 122:
```suggestion
		err = ioutil.WriteFile(privateKeyFile, file, 0600)
		defer 		os.Remove(privateKeyFile)
		if err != nil {
			log.Fatalf("Unable to write sealed-secrets-key to disk: %v", err)
		}
		kubeseal := p.GetBinary("kubeseal")
		flagString := strings.Join(flags, " ")
		argString := strings.Join(args, " ")
		if err := kubeseal(flagString + argString); err != nil {
			log.Fatalf("failed to run kubeseal: %v", err)
		}
```

cmd/unseal.go
```diff
@@ -0,0 +1,129 @@
...
+			log.Fatalf("Unable to resolve file path for %s: %v", privateKeyFile, err)
+		}
+		flags := []string{
+			"--recovery-unseal",
+			getFlagString("format", cmd),
+			getFlagBool("allow-empty-data", cmd),
+			fmt.Sprintf("--recovery-private-key %s", filePath),
+		}
+		if !p.SealedSecrets.Disabled && p.SealedSecrets.Certificate != nil {
+			if p.SealedSecrets.Certificate.Cert != "" {
```
moshloop, line 87: If we dont't have a copy of the cert shouldn't we fail fast ?

cmd/seal.go
```diff
@@ -0,0 +1,102 @@
...
+		argString := strings.Join(args, " ")
+		if err := kubeseal(flagString + argString); err != nil {
+			log.Fatalf("failed to run kubeseal: %v", err)
+		}
+	},
+}
+
+func init() {
+	Seal.Flags().StringP("format", "o", "yaml", "Output format for sealed secret. Either json or yaml")
+	Seal.Flags().StringP("sealed-secret-file", "w", "", "Sealed-secret (output) file")
```
moshloop, line 94: This shouldn't be necessary - we always output to stdout

cmd/seal.go
```diff
@@ -0,0 +1,102 @@
...
+	Seal.Flags().Bool("fetch-cert", false, "Write certificate to stdout. Useful for later use with --cert")
```
moshloop, line 95: Why do we need this ?

cmd/unseal.go
```diff
@@ -0,0 +1,129 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		p := getPlatform(cmd)
+		filePath, err := filepath.Abs(privateKeyFile)
+		if err != nil {
+			log.Fatalf("Unable to resolve file path for %s: %v", privateKeyFile, err)
+		}
+		flags := []string{
+			"--recovery-unseal",
+			getFlagString("format", cmd),
+			getFlagBool("allow-empty-data", cmd),
```
moshloop, line 83: Why do we need this ?

cmd/seal.go
```diff
@@ -0,0 +1,102 @@
...
+	Seal.Flags().Bool("validate", false, "Validate that the sealed secret can be decrypted")
+	Seal.Flags().String("merge-into", "", "Merge items from secret into an existing sealed secret file, updating the file in-place instead of writing to stdout.")
+	Seal.Flags().Bool("raw", false, "Encrypt a raw value passed via the --from-* flags instead of the whole secret object")
```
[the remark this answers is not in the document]
paddatrapper: merge-into was helpful several times during the BCB setup when we needed to append a secret to an existing sealed-secret. The others I'll remove

PR #932, opened 2021-07-10
feat: exclude specific alert rules

### LINKED ISSUE #911: Exclude alerts
e.g. 

```yaml
monitoring:
    excludeAlerts:
      - ExcessivePodCPURatio
```

teodor-pripoae:
Fixes #911


moshloop (changes requested):
Can we create a Makefile for these types that pulls from  upstream so that we can maintain them better ?

moshloop:
@teodor-pripoae  - Can we create a Makefile for these types that pulls from upstream so that we can maintain them better ?

moshloop:
merged into https://github.com/flanksource/karina/pull/938

PR #949, opened 2021-09-07
feat: add github auth config to dex

BrendanGalloway:
### Description

- Add github oauth struct to dex struct
- more closely align templating logic with file layout
- add github secret creation to dex install

### Breaking Change

- [ ] Yes
- [x] No

### DIFF
--- a/manifests/dex.cfg
+++ b/manifests/dex.cfg
@@ -14,7 +14,7 @@ frontend:
    issuerUrl: "{{.brand.url}}"
    logoUrl: "{{.brand.logo}}"
    {{ end }}
-logger
+logger:
   level: "debug"
 oauth2:
   skipApprovalScreen: true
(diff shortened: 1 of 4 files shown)


manifests/dex.cfg
```diff
@@ -14,7 +14,7 @@ frontend:
    issuerUrl: "{{.brand.url}}"
    logoUrl: "{{.brand.logo}}"
    {{ end }}
-logger:
+logger
```
paddatrapper, line 17: Was this intentional?

manifests/dex.cfg
```diff
@@ -32,8 +32,30 @@ connectors:
...
+      {{ end }}
+  {{ end }}
+  {{ if index .dex "github" }}
+    - type: github
+      id: github
+      name: Github
+      config:
+        clientID: $GITHUB_CLIENT_ID
+        clientSecret: $GITHUB_CLIENT_SECRET
+        redirectURI: http://127.0.0.1:5556/dex/callback
```
moshloop, line 44: I don't think this redirectURI will work

PR #953, opened 2021-09-10
Template operator resources

BrendanGalloway:
### Description

Increase resources in template operator manifest

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

Generated manifests, fixed faulty log exporter patch


manifests/upstream/template-operator/base/kustomization.yaml
```diff
@@ -24,7 +24,7 @@ patchesStrategicMerge:
               resources:
                 limits:
                   cpu: 150m
-                  memory: 200Mi
+                  memory: 2Gi
                 requests:
                   cpu: 100m
-                  memory: 150Mi
+                  memory: 1Gi
```
moshloop, line 30: I don't think we can do this by default - a 200Mi request and 1Gi limit should be sufficient, cpu limit can also be bumped to 500m

BrendanGalloway: ok.  I thought these were the numbers you mentioned in the meeting, but I can adjust them

PR #985, opened 2021-12-21
chore: bump canary-checker version

paddatrapper:
### Description

Bumps canary-checker to v0.38.67

### Breaking Change

- [ ] Yes
- [x] No

### DIFF
--- a/manifests/upstream/canary-checker/base/kustomization.yaml
+++ b/manifests/upstream/canary-checker/base/kustomization.yaml
@@ -1,15 +1,16 @@
 resources:
-  - https://raw.githubusercontent.com/flanksource/canary-checker/v0.38.30/config/deploy/base.yaml
+  - https://raw.githubusercontent.com/flanksource/canary-checker/v0.38.67/config/deploy/base.yaml
 images:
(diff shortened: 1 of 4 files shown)


manifests/upstream/canary-checker/base/kustomization.yaml
```diff
@@ -1,15 +1,16 @@
...
+    newTag: '{{.canaryChecker.version | default "v0.38.67" }}'
+namespace: platform-system
 patchesStrategicMerge:
   - |-
     apiVersion: apps/v1
     kind: Deployment
     metadata:
       name: canary-checker
-      namespace: platform-system
+      namespace: canary-checker
```
moshloop, line 13: I don't think we want to change the default deployment from platform-system to canary-checker

moshloop (changes requested):

manifests/upstream/canary-checker/base/kustomization.yaml
paddatrapper: We're not. It's just that the patch is applied before the namespace is changed in kustomize, so for these to apply, they need to match the original namespace

PR #995, opened 2022-01-13
1.22 compat

BrendanGalloway:
### Description

Update manifests containing deprecated apiversions 

### Dependencies

NA

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

WIP - triggering sweep of automated tests


manifests/argo-rollouts.yaml
```diff
@@ -287,9 +334,29 @@ spec:
         app.kubernetes.io/name: argo-rollouts
     spec:
       containers:
-      - image: docker.io/argoproj/argo-rollouts:{{.argoRollouts.version}}
+      - image: quay.io/argoproj/argo-rollouts:latest
```
paddatrapper, line 337: Looks like the kustomize image needs updating to reflect quay.io so that the tag gets patched

manifests/upstream/calico/crd/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+bases:
```
moshloop, line 1: I think this is a little weird to install calico from cloud-provider-gcp, I would add a comment as to why

moshloop (changes requested):
Can we also adjust the test matrix to add 1.22? And then also move kpack/quarantine/cicd tests that always fail to a nightly build ?

manifests/upstream/calico/crd/kustomization.yaml
BrendanGalloway: I couldn't find versioned static manifests published by Calico themselves - The calico docs point to unversioned artifacts https://projectcalico.docs.tigera.io/getting-started/kubernetes/quickstart and the calico github only has the inputs to generate the artefacts rather than the artefacts themselves.  If there's a place where they are actually available that I missed please let me know.

moshloop: Thats fine - Just put that message as a comment in the code

PR #1005, opened 2022-03-29
Feat: add alerting hook for nginx specific alerts

BrendanGalloway:
### Description

Adds a new manifest and corresponding install phase actions to deploy alerts specific to the nginx ingress controller

### Dependencies

NA

### Breaking Change

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/phases/nginx/install.go
+++ b/pkg/phases/nginx/install.go
@@ -133,6 +133,12 @@ func Install(platform *platform.Platform) error {
 	if err := platform.ApplySpecs(v1.NamespaceAll, "nginx.yaml"); err != nil {
 		return err
 	}
+	// nginx alerts live in a separate manifest as alert syntax often cannot be templated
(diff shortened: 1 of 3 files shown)


pkg/phases/nginx/install.go
```diff
@@ -133,6 +133,12 @@ func Install(platform *platform.Platform) error {
 	if err := platform.ApplySpecs(v1.NamespaceAll, "nginx.yaml"); err != nil {
 		return err
 	}
+	// nginx alerts live in a separate manifest as alert syntax often cannot be templated
+	if !platform.Monitoring.IsDisabled() {
```
moshloop, line 137: but in this case it isn't an issue ?

BrendanGalloway: Not in this case no, but if we ever add an alert that needs a filter then it would be, so I figured it would be best to start with everything in the state that can handle it, rather than needing it to be reworked later.

PR #1006, opened 2022-03-30
chore: update canary-checker

paddatrapper:


manifests/canary-checker.yaml
```diff
@@ -152,8 +156,8 @@ spec:
         com.flanksource.infra.logs/processors.1.drop_event.when.contains.message: '[pod/canary] request completed with 503, expected [200 201 202], retrying'
         com.flanksource.infra.logs/processors.2.drop_event.when.contains.message: Requeue reconciliation
         com.flanksource.infra.logs/processors.3.drop_event.when.contains.message: Successfully Reconciled
-        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
```
moshloop, line 155: Don't we need to keep this ?

manifests/upstream/canary-checker/helm/values.yaml
```diff
@@ -0,0 +1,14 @@
+image:
+  tag: 'latest'
```
moshloop, line 2: should thos be .canaryChecker.version ?

paddatrapper: It's patched to that via kustomize. I couldn't place it here because the helm chart validation refused to accept it

manifests/canary-checker.yaml
paddatrapper: Yes, fixed

BrendanGalloway: it moved to the top of the annotation list.  I does need to be templated to be false if local-path storage is specified for the DB though, else we will lose the data...

manifests/canary-checker.yaml
```diff
@@ -228,13 +242,18 @@ spec:
...
+      app.kubernetes.io/instance: canary-checker
+      app.kubernetes.io/name: canary-checker
       control-plane: canary-checker
 ---
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
-    kubernetes.io/tls-acme: "true"
+    kubernetes.io/ingress.class: nginx
```
BrendanGalloway, line 253: Is this patched in?  It should not be the default in the base template...and I don't see it in the values.yaml

manifests/canary-checker.yaml
```diff
@@ -228,13 +242,18 @@ spec:
...
     matchLabels:
+      app.kubernetes.io/instance: canary-checker
+      app.kubernetes.io/name: canary-checker
       control-plane: canary-checker
 ---
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
-    kubernetes.io/tls-acme: "true"
```
BrendanGalloway, line 237: this needs to stay true?  OR are we currently patching it in later along with oauth instructions?

manifests/canary-checker.yaml
paddatrapper: It's the default in the canary-checker helm chart values. It probably shouldn't be default there and be left to downstream consumers to set

manifests/canary-checker.yaml
paddatrapper: Fixed

manifests/canary-checker.yaml
paddatrapper: https://github.com/flanksource/canary-checker/pull/443

paddatrapper: This has now been dropped upstream and updated here

manifests/canary-checker.yaml
paddatrapper: Added the condition

PR #1009, opened 2022-04-06
fix: use correct rbac for grafana controller

BrendanGalloway:
### Description

It looks like the incorrect set of rbac got linked in to the grafana operator manifest during the 1.22 compat bump.  Fixed by adding in what is actually in the linked doc upstream

### Dependencies
NA

### Breaking Change

- [ ] Yes
- [x] No


moshloop:
I am not sure this is going to work ? the names no longer match or is their a rename happening somewhere ?

moshloop (approved):

PR #1014, opened 2022-04-26
chore: bump kube-rbac-proxy versions

BrendanGalloway:
### Description

Older kube-rbac-proxy images use the deprecated authentication/v1beta api call rather than v1.  Newer image version are needed for kubernetes 1.22

### Dependencies

NA

### Breaking Change

- [ ] Yes
- [x] No

### Testing Notes

Verified Prometheus is able scrape node-exporter pods with the new sidecar image

### DIFF
--- a/manifests/monitoring/node-exporter.yaml
+++ b/manifests/monitoring/node-exporter.yaml
@@ -86,7 +86,7 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: status.podIP
-          image: bitname/kube-rbac-proxy:v0.12.0
+          image: quay.io/brancz/kube-rbac-proxy:v0.11.0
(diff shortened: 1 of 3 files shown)


manifests/monitoring/node-exporter.yaml
```diff
@@ -86,7 +86,7 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: status.podIP
-          image: quay.io/coreos/kube-rbac-proxy:v0.4.1
+          image: bitname/kube-rbac-proxy:v0.12.0
```
moshloop, line 89:
```suggestion
          image: quay.io/brancz/kube-rbac-proxy:v0.11.0
```

PR #1033, opened 2022-06-29
feat: add support for grafana generic oauth login

paddatrapper:


moshloop:
Maybe its better we just patch this ?

paddatrapper:
Ideally it would be great if we could just dump the yaml from the karina config into here, then it could be generic and not depend on support for a specific OAuth provider. However, I didn't manage to find a clean way to do that

PR #1049, opened 2022-10-04
fix: correctly alert when certificates are close to expiring

paddatrapper:


BrendanGalloway (approved):

BrendanGalloway (changes requested):
Let's also bump the latch time to 10min?  We've seen zerossl takes a bit longer to issue certs than letsencrypt and getting the alert 28.99 days ahead of expiry instead of 29 is probably ok

BrendanGalloway (approved):

PR #1071, opened 2023-03-10
added base for sealedsecrets

edeediong:
### Description

### Dependencies

- _Ex. Other PRs._

### Breaking Change

- [ ] Yes
- [ ] No

### Testing Notes

**What I did:**

### **Links**


manifests/upstream/sealed-secrets/base/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources: 
+- https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.19.5/controller.yaml
```
BrendanGalloway, line 2: this will additionally need an 'images' directive to insert templating directives to pass version info through from the karina file.  Also, if the deployment features an emptyDir{} volume mount for things like logs you'll want to patch in the safe-to-evict annotation so that it doesn't break autoscaling.

once you've done that, you can try:
```
cd manifests
make sealed-secrets.yaml
```
and check that the manifest has been generated and the templates were inserted., then add the updated manifests to the PR

BrendanGalloway (changes requested):

BrendanGalloway (changes requested):
I did not realise the CRDs were in the upstream file, we probably want to remove those here since we apply crds explicitly at an earlier phase via a crds manfiest.  `$patch delete` in the kustomize might do the trick for removal.  Also please update the crd kustomize version ref to match the version you're pulling for the base spec.

manifests/sealed-secrets.yaml
```diff
@@ -1,163 +1,67 @@
...
 kind: ServiceAccount
 metadata:
-  annotations: {}
+  annotations:
+    cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
   labels:
     name: sealed-secrets-controller
   name: sealed-secrets-controller
-  namespace: sealed-secrets
+  namespace: kube-system
```
paddatrapper, line 10: You'll need to set the namespace to `sealed-secrets` in kustomize

paddatrapper (changes requested):

manifests/upstream/sealed-secrets/base/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
...
```
BrendanGalloway, line 2: this is v0.I9.5 - is the I correct?

manifests/upstream/sealed-secrets/crd/kustomization.yaml
```diff
@@ -1,2 +1,2 @@
 resources:
-  - https://raw.githubusercontent.com/bitnami-labs/sealed-secrets/v0.17.1/helm/sealed-secrets/crds/sealedsecret-crd.yaml
\ No newline at end of file
+  - https://raw.githubusercontent.com/bitnami-labs/sealed-secrets/v0.17.1/helm/sealed-secrets/crds/sealedsecret-crd.yaml
```
BrendanGalloway, line 2: Set the version to the same as the base version (0.19.5?) and then also make crds/sealed-secrets.yaml

BrendanGalloway (changes requested):

PR #1083, opened 2023-08-24
fix: add flag to disable kube-state-metrics deployment

dabasvibhor:


moshloop:
@dabasvibhor why do we want to remove this? If anything just don't deploy it monitoring=false

